PR #1, opened 2024-02-21
Update steamodded_injector.py to remove suspicious discord links

NanashiTheNameless:
Update steamodded_injector.py to remove suspicious discord links

Did you know the 7-zip "installer" can be opened as a .zip archive to get a portable release? 
No?
Now you do!

### DIFF
--- a/injector/steamodded_injector.py
+++ b/injector/steamodded_injector.py
@@ -61,7 +61,7 @@ def modify_main_lua(main_lua_path, base_dir, directories):
 print("Starting the process...")
 
 # URL to download the LuaJIT decompiler
-luajit_decompiler_url = "https://cdn.discordapp.com/attachments/485484159603572757/1185701932707369111/luajit-decompiler-v2.exe?ex=659091fa&is=657e1cfa&hm=74df61cc183f19dda8a9a4ee079b659543bbde6e387b8b5d624ac51b6a92fed6&"
+#luajit_decompiler_url = "https://cdn.discordapp.com/attachments/485484159603572757/1185701932707369111/luajit-decompiler-v2.exe?ex=659091fa&is=657e1cfa&hm=74df61cc183f19dda8a9a4ee079b659543bbde6e387b8b5d624ac51b6a92fed6&"
 
 # Temporary directory for operations
 with tempfile.TemporaryDirectory() as decompiler_dir:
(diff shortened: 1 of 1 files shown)


SamuelDev (approved):

Steamopollys:
In fact, I did not knew that !

I will test it and merge it before deploying the next Release.

Steamopollys:
Hmmmm. It's indeed an SFX .zip file but python is not happy about it:

`zipfile.BadZipFile: File is not a zip file`

I will try to find a solution to this, but it might need a bit more time than expected.

Steamopollys:
Wait, It's because there is a type in the MR, let me patch it.

injector/steamodded_injector.py
```diff
@@ -76,13 +76,13 @@ def modify_main_lua(main_lua_path, base_dir, directories):
...
     # URL to download the 7-Zip suite
-    seven_zip_url = "https://cdn.discordapp.com/attachments/485484159603572757/1181530870029504514/7-Zip.zip?ex=6581655f&is=656ef05f&hm=3dfd3e5a4936b0a50d7a5b3f2dd36c30032940b6fb6715e2399602d3be68ce0e"
+    seven_zip_url = "https://7-zip.org/a/7z2301-x64.exe"
 
     # Temporary directory for 7-Zip suite
     with tempfile.TemporaryDirectory() as seven_zip_dir:
         print("Downloading and extracting 7-Zip suite...")
-        download_file(seven_zip_url, os.path.join(seven_zip_dir, "7-Zip.zip"))
-        with zipfile.ZipFile(os.path.join(seven_zip_dir, "7-Zip.zip"), 'r') as zip_ref:
+        download_file(seven_zip_url, os.path.join(seven_zip_dir, "7z2301-x64.zip"))
```
Steamopollys, line 84: 7z2301-x64.exe

injector/steamodded_injector.py
```diff
@@ -76,13 +76,13 @@ def modify_main_lua(main_lua_path, base_dir, directories):
...
+        with zipfile.ZipFile(os.path.join(seven_zip_dir, "7z2301-x64.zip"), 'r') as zip_ref:
```
Steamopollys, line 85: 7z2301-x64.exe

Steamopollys (changes requested):
7z2301-x64.exe

Steamopollys:
> is there any reason to not just include a copy of 7-zip statically?

You mean in the executable ?

Steamopollys:
Apparently it was something else, Python is not able to handle it like a .zip for whatever reason. I will work on it over the next few days.

Steamopollys:
> > > is there any reason to not just include a copy of 7-zip statically?
> > 
> > 
> > You mean in the executable ?
> 
> yeah, is there any reason not to just include 7-zip instead of downloading it?

Not really, simply the binary size I would say. Bundling 7Zip directly inside the .exe can also be a solution.

ebertin42:
> > > > > is there any reason to not just include a copy of 7-zip statically?
> > > > 
> > > > 
> > > > You mean in the executable ?
> > > 
> > > 
> > > yeah, is there any reason not to just include 7-zip instead of downloading it?
> > 
> > 
> > Not really, simply the binary size I would say. Bundling 7Zip directly inside the .exe can also be a solution.
> 
> Given the current size so far, I don't particularly see a reason to avoid packing it inside for the sake of user experience

It can be a solution. If using the self extracting .exe is to complicated, I will include the .zip in the Nuitka bundle.

Steamopollys:
Everything looks fine to me, but I would like to keep the old file structure of the project. It's possible to change the target path of Nuitka to go in the directory above, so I would prefer this option.

Steamopollys (approved):

Steamopollys:
There is also some issues with TempDirectory scope that I would like to review before merging.

PR #6, opened 2024-02-25
Add hook for custom mod menus

adambennett:
Add a hook to allow mods to define their own UI buttons after the mod button has been clicked open

Planning to use this for a Custom Deck creation mod I've been working on. I'd like to have a menu be openable from within my mod's menu so as not to pollute the main game menu with more buttons.

I'm new to Lua, so if this isn't the best way to accomplish what I want please feel free to suggest a better approach!

Also fixes a few typos and updates the gitignore to ignore some IntelliJ files


Steamopollys:
Hello,

A custom UI for Mods is already in preparation, but in a different form. I want to add options selections for the mods and being able to enable / disable them.

It will be a little different than this one since it's providing the base structure to "standardize" experience between mods.

However, I agree that the possibility to add a fully customizable menu will be interesting, but I would hook it over the base function.

I let this PR open to see if we can implement it with what I explained before!

adambennett:
Ok thanks Steamo. Any idea of ETA when this will be usable? My deck creator is almost complete, and I really just need a way to add one button on the inner menu!

Steamopollys:
I wanted to add the JokerAPI first before going over this subject.

I think the best idea would be to merge it now but I would change it in some versions to evolve to the new system. It would mean that you will have to update your Mod to match the new system.

If it's ok for you I will review it and merge it asap.

adambennett:
Yes that will work for me. I'll keep an eye out for any updates involving this in the future.

I'd be happy to help contribute to the design improvements here as well if you'd like, but I need to understand more about how the UI code works in the game first. This is mostly what I'm working on now, so maybe I will be able to come up with some ideas that could help organize this API better soon.

Appreciate your time on this! Thanks!

Steamopollys:
Alright, you will be able to test the build coming from: https://github.com/Steamopollys/Steamodded/actions/runs/8039277869/job/21956121462

I will do an official release later when I do have more content to add.

PR #10, opened 2024-02-26
Extract injector body to main function, to allow importing

eutro:
This PR introduces a `main` function and calls it from an `if __name__ == "__main__"` block. The purpose of this is to ensure that the functions can be imported from other scripts without running the program. This is helpful for testing, or for other tools that may want to cooperatively use the injector.

While writing this I also wanted to make sure I didn't break anything. Unfortunately, the script didn't even work on my machine (Linux). Thus I made a handful of minor semantic changes as well, both to fix this, and to make actually finding such errors easier.

All the changes this PR makes:

- Introduce `main` function, as well as `download_seven_zip` and `modify_game_sources`.
- Changed 7zip command to use `u` instead of `a`, since files are only being updated in the archive
  - Question: 7zip is being downloaded and extracted even the system version is used, could this be fixed?
- Use the parent of the script's directory for `base_dir`, to ensure that the files are actually found when running from source
- Raise exceptions in `except IOError` blocks, because if there was an error then the process likely didn't complete successfully.
- Raise an exception if the `----MOD CORE----` comment is found in the code, to ensure that we don't inject multiple times silently.
- Remove `shutil` import, it is no longer used

### DIFF
--- a/injector/steamodded_injector.py
+++ b/injector/steamodded_injector.py
@@ -25,7 +25,7 @@ def merge_directory_contents(directory_path):
     directory_content = ""
     if os.path.exists(directory_path):
         print(f"Processing directory: {directory_path}")
-        for file_name in os.listdir(directory_path):
+        for file_name in sorted(os.listdir(directory_path)):
             if file_name.endswith(".lua"):
                 file_path = os.path.join(directory_path, file_name)
                 try:
--- a/injector/steamodded_injector.py
+++ b/injector/steamodded_injector.py
@@ -1,11 +1,11 @@
-import subprocess
 import os
+import platform
+import subprocess
 import sys
 import tempfile
 import zipfile
-import shutil
-import platform
 import requests
+
 
 def download_file(url, output_path):
     response = requests.get(url, stream=True)
@@ -35,8 +35,10 @@ def merge_directory_contents(directory_path):
                         print(f"Appended {file_name} to the directory content")
                 except IOError as e:
                     print(f"Error reading {file_path}: {e}")
+                    raise RuntimeError()
     else:
         print(f"Directory not found: {directory_path}")
+        raise RuntimeError()
     return directory_content
 
 
@@ -48,7 +50,11 @@ def modify_main_lua(main_lua_path, base_dir, directories):
             main_lua_content = file.read()
     except IOError as e:
         print(f"Error reading {main_lua_path}: {e}")
-        return
+        raise RuntimeError()
+
+    if "----MOD CORE----" in main_lua_content:
+        print(f"Mod loader already present in {main_lua_path}, aborting")
+        raise RuntimeError()
 
     for directory in directories:
         directory_path = os.path.join(base_dir, directory)
@@ -61,6 +67,7 @@ def modify_main_lua(main_lua_path, base_dir, directories):
             file.write(main_lua_content)
     except IOError as e:
         print(f"Error writing to {main_lua_path}: {e}")
+        raise RuntimeError()
 
 
 def modify_game_lua(game_lua_path):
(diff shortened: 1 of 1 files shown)


Steamopollys:
Hello,

That's a lot of good changes. The injector became messier with the last updates and it will be a good thing  to clean it a little bit.

For the 7zip questions, what do you mean ?

The `----MOD CORE----` is also something that I wanted to add in the future so this is a welcome addition.

For `base_dir`, I will need to test a little bit. Nuitka can some time mess a lot with the path.

I will test it and merge it now before deploying the next update.


However that's the last "big" update to the Python version I will accept. AV false flags are becoming a true issue and I want to switch to Rust for a new version in a near future.

eutro:
The 7zip distribution is extracted even on Linux where it's not used.

I believe I saw someone on Discord having the same path issue. It shouldn't affect the Nuitka build if the "frozen" branch is properly taken? Though I haven't tested it on a Windows machine.

Steamopollys:
Yep, looks like the changes for the `base_dir` is messing up the Nuitka env:

```
Modifying C:\Users\X\AppData\Local\Temp\tmp6rrc2_7o\main.lua with files from ['core', 'debug', 'loader'] in C:\Users\X\AppData\Local\Temp\ONEFIL~3\..
Looking for directory: C:\Users\X\AppData\Local\Temp\ONEFIL~3\..\core
Directory not found: C:\Users\X\AppData\Local\Temp\ONEFIL~3\..\core
```

If you can revert this part.

Steamopollys:
> The 7zip distribution is extracted even on Linux where it's not used.
> 
> I believe I saw someone on Discord having the same path issue. It shouldn't affect the Nuitka build if the "frozen" branch is properly taken? Though I haven't tested it on a Windows machine.

On Windows that's this version of 7zip that is used, not the system one.

Steamopollys:
`base_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..")`

The repo structure as changed on this part, I assume you will also need to modify it.

Steamopollys:
Wait, it was part on the error for windows. This part of the code: 

```
if getattr(sys, "frozen", False):
    # Running in a PyInstaller or Nuitka bundle
    base_dir = os.path.dirname(sys.executable)
else:
    # Running in a normal Python environment
    base_dir = os.path.dirname(os.path.abspath(__file__))
```

Is apparently not working anymore on a windows. I assume that Nuitka is closer to a classic Python env than Pyinstaller was. I won't delete it for now because I know that some user have issues with Nuitka and are building it with Pyinstaller instead. 

Everything is working on my side with this little change. Sorry for the oversight, I was testing with the updates file structure.

I let you do the modifications on your side and I will them merge it into the main repo.

injector/steamodded_injector.py
```diff
@@ -85,127 +92,145 @@ def modify_game_lua(game_lua_path):
...
+    subprocess.run([seven_zip_command, "x", f"-o{temp_dir.name}", sfx_archive_path], check=True)
+    print("Extraction complete.")
+
+    # Determine the base directory (where the .exe is located)
+    if getattr(sys, "frozen", False):
+        # Running in a PyInstaller or Nuitka bundle
+        base_dir = os.path.dirname(sys.executable)
+    else:
+        # Running in a normal Python environment
+        base_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..")
```
Steamopollys, line 208: `base_dir = os.path.dirname(os.path.abspath(__file__))`

To match new file structure

injector/steamodded_injector.py
```diff
@@ -85,127 +92,145 @@ def modify_game_lua(game_lua_path):
...
+    # Temporary directory for extraction and modification
+    temp_dir = tempfile.TemporaryDirectory()
+    print(temp_dir.name)
+    # Extract the SFX archive
+    subprocess.run([seven_zip_command, "x", f"-o{temp_dir.name}", sfx_archive_path], check=True)
+    print("Extraction complete.")
+
+    # Determine the base directory (where the .exe is located)
+    if getattr(sys, "frozen", False):
+        # Running in a PyInstaller or Nuitka bundle
```
Steamopollys, line 204: `# Running in a PyInstaller bundle`

injector/steamodded_injector.py
```diff
@@ -85,127 +92,145 @@ def modify_game_lua(game_lua_path):
...
+        base_dir = os.path.dirname(sys.executable)
+    else:
+        # Running in a normal Python environment
```
Steamopollys, line 207: `# Running in a normal Python environment or Nuitka bundle`

Steamopollys (changes requested):

PR #11, opened 2024-02-27
Added a function wrapper to set_sprites to allow Jokers to have custom atlases without jank

GoldenEpsilon:
Without this change the "atlas" variable only does something the SECOND time set_sprites is called, and this code is only needed to exist once generically alongside being generic itself so there is no real reason not to have it within steamodded itself

### DIFF
--- a/core/sprite.lua
+++ b/core/sprite.lua
@@ -166,7 +166,7 @@ local set_spritesref = Card.set_sprites
 function Card:set_sprites(_center, _front)
     if _center then 
         if _center.set then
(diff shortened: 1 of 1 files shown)


Steamopollys:
Hello, thanks for the PR I will test it.

If it works correctly, I will appear in the JokerAPI release.

Steamopollys:
The Joker API entered PRE Release phase, I will merge this after I'm sure everything is good on the JAPI part.

Steamopollys:
I will also edit the Sprite object to better understand custom sprite.

PR #29, opened 2024-03-08
make joker sorting consistent

notmario:
make joker sorting use ipairs() instead of pairs() and sort jokers before injecting


Steamopollys:
Hello, this change cause some... issues with the jokers. I'm not sure exactly why, I will investigate.

Aurelius7309:
Hello, I've implemented a different approach to the same problem you tried to fix in my recent overhaul of center APIs, also considering the loading order priority system of mods.

PR #30, opened 2024-03-09
Update the Linux install command

Optissimum:
Updates the command to correctly hand off the Balatro executable as the SFX file that is being edited.


Steamopollys:
Forgot to update it. I will merge it.

PR #34, opened 2024-03-17
Playing Card API bug fixes

Aurelius7309:
Fixes: 
- Fix a bug with get_straight deviating from the base game's behavior, not allowing five card straights with four fingers
- Fix an issue with is_face not accounting for stone cards properly, causing stone cards that were face cards originally to be considered face cards
- Address an issue where added playing cards are unloaded on profile/language change.

Known issues:
- The playing card API currently does not offer multi-language support, and any information inserted regarding new playing cards is erased when changing languages. This will be implemented at a later date.

### DIFF
--- a/core/suit.lua
+++ b/core/suit.lua
@@ -7,29 +7,25 @@ SMODS.Card.SUITS = {
 		name = 'Hearts',
 		prefix = 'H',
 		suit_nominal = 0.03,
-        ui_pos = { x = 0, y = 1 },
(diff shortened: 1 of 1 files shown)


Aurelius7309:
New commit fixes an issue with cards not being reachable starting from Aces in get_straight. Also I removed the upper card limits in get_flush and get_straight, as this is an irrelevant case in the base game, and removing it allows mods to check if the entirety of cards held in hand can make specific hands.

PR #42, opened 2024-03-22
Add joker ability UI support

GauntletGames-2086:
Every custom joker mod so far has to append the `generate_UIBox_ability_table` function so the ability UI box for jokers displays the variables in their config. I suggest that at some point there is proper support for this in Steamodded. 
To make it work, add this into your mod code: 
```
SMODS.Jokers.j_joker_slug.UIBox_info = function(self)
	return {} --Variables you want to localize go here
end
```
Note: Code does have a bug. The joker's name will be displayed on the main menu screen in the top left corner when highlighted. Couldn't figure out the cause, so had to leave it in.

### DIFF
--- a/core/joker.lua
+++ b/core/joker.lua
@@ -158,5 +158,52 @@ function SMODS.end_calculate_context(c)
     return false
 end
 
+local generate_UIBox_ability_tableref = Card.generate_UIBox_ability_table;
+function Card:generate_UIBox_ability_table()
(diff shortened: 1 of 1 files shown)


Aurelius7309:
Thanks for this suggestion. I've applied some optimizations to your code and incorporated it alongside some other improvements I was making.

GauntletGames-2086:
0.8.1 has the optimized version, now resolved

PR #54, opened 2024-03-28
Improve debug console

jubnl:
Improve debug console with the following functionnality:
- Resize window
- CTRL + F to search something in the console
- 3 search mode
1. Normal mode
2. Regex mode
3. Match case mode
- Press Enter to search for the next occurence
- Clear logs button

in search bar :
Ctrl + backspace to remove an entire word


Aurelius7309:
Hello, I quickly tested this and it looks like it's working perfectly! Merging shortly

PR #56, opened 2024-03-29
Improve logging

jubnl:
Log levels implemented:
- TRACE
- DEBUG
- INFO
- WARN
- ERROR
- FATAL

Also made sure that if 2 logs are sent and received at the same time, they will appear on 2 different lines.

The logs are also timestamped.

In the different sendLEVELMessage, you now can add a logger (just a string) to identify what mod, or event different type of logger to identify easily from where the log comes from.

usage example :

```lua
sendInfoMessage("My Custom Message", "MyLoggerName")
```

will print out in the console :
```text
2024-03-29 15:18:18 :: INFO  :: MyLoggerName :: My Custom Message
```

The logger name is optional and will default to "DefaultLogger" if the argument is not passed.

I'm currently working on a way to exploint the new logger format console side. Specifically 2 filters that will only show logs from a specific level (or from a specific level and above) and filter out only logs from a specific logger

### DIFF
--- a/core/StackTracePlus.lua
+++ b/core/StackTracePlus.lua
@@ -455,7 +455,7 @@ function injectStackTrace()
 
     local crashHandler = love.errhand
     function love.errhand(msg)
-        sendDebugMessage("Oops! The game crashed\n" .. STP.stacktrace(msg))
+        sendErrorMessage("Oops! The game crashed\n" .. STP.stacktrace(msg))
         if _RELEASE_MODE then
             msg = STP.stacktrace(msg .. "\n", 3) -- The game shows the stack for me if release mode is false
         end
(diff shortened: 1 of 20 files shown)


Aurelius7309:
Am I correct in thinking this is incomplete as of now? This seems like a great addition!

jubnl:
The is the 1st advancement before having a proper console. It is missing the console part, which I'm working on at the moment. This part can already be implemented in the main repo imo, so modders (and the Steamodded team) can already utilize the logging part. At the moment, the search part combined with the new logger will already help a lot the devs!

I'm rewriting the console from scratch, once I got it working, I'll push it and create a new pull request if that's ok with you

PR #58, opened 2024-03-29
feat: add in logic for adding seals into the pack opening

mwithington:
This fix is needed to override the original code for the `card.open` function so seals can be added to cards when a pack is opened. Currently only the original seals will be added to cards so even though the seal is added to the game it will not appear in runs. 

This is probably not the correct way of implementing this and I am happy to change it to a new file called `pack.lua` but I figured its better to get some eyes on it first.

### DIFF
--- a/core/seal.lua
+++ b/core/seal.lua
@@ -74,3 +74,138 @@ function get_badge_colour(key)
 	end
     return fromRef
 end
+
+-- Orginally from @axbolduc
+-- Add the seal to be randomly generated in standard packs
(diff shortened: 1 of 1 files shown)


Aurelius7309:
It would be better to avoid destructive patches like this one. I hope to be able to resolve this more gracefully using lovely once they have multiline pattern recognition.

Aurelius7309:
This has been implemented using lovely and will be included in the upcoming major release.

PR #68, opened 2024-04-05
Print on debug msg

WilsontheWolf:
This PR makes it where the debug methods also print to the lua console. This is useful for cases where there is a debug console (such as with lovely), where one does not need to run steamodded's debug console.


jubnl (approved):
lgtm

PR #70, opened 2024-04-07
Fully custom error handler

WilsontheWolf:
This PR modifies the better stack trace, to use its own implementation for rendering the error. This provides the following benefits
- No more hacks to show the stack trace (and it has consistent output whether or not the user has debug mode enabled)
- Won't send crash logs to localthunk
- Log can be copied with ctrl + c
- Log can be scrolled if its too big for the screen
- Now shows some other info, which might help in debugging. Namely: Balatro Version, Steamodded Version, Lovely Version (if applicable), list of steamodded mods.
- Much easier to add more functionality in the future, if we so desire

I've also made a new api. Mod's can now set SMOD.DebugInfo.modid to either a string, or table of strings, and the crash handler will display them beneath the mod. This is designed for mods to show whatever they want. This can include things like version numbers, internal states, and anything else they might find useful to show.

### DIFF
--- a/core/StackTracePlus.lua
+++ b/core/StackTracePlus.lua
@@ -452,10 +452,29 @@ function getDebugInfoForCrash()
     if lovely_success then
         info = info .. "\nLovely Version: " .. lovely.version
     end
-    info = info .. "\nSteamodded Mods:"
-    for k, v in pairs(SMODS.MODS) do
-        info = info .. "\n    " .. k .. ": " .. v.name .. " By " .. concatAuthors(v.author) .. " [ID: " .. v.id ..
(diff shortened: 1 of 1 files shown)


WilsontheWolf:
Video Demo:

https://github.com/Steamopollys/Steamodded/assets/33164598/b10162a9-db30-46d3-9221-cf1bdd3ab43a

WilsontheWolf:
I was messing around and I realized I could make it where mods can provide some debug info. 
For example
image
makes
image
and 
image
makes
image
Maybe the api would need to be defined differently, but I think this would be super helpful for mod developers

WilsontheWolf:
Sorry I keep adding stuff

WilsontheWolf:
Is there any update on this being reviewed/merged?

PR #71, opened 2024-04-07
Refactor injector, auto-create mods directory

### LINKED ISSUE #23: MacOS modding currently broken
The code seems to be written to only accept exe files. The resources directory in the .App format is different than what the code accepts

/Library/Application Support/Steam/steamapps/common/Balatro/Balatro.app/Contents/Resources

Balatro.love
GameIcon.icns
libsteam_api.dylib
license.txt
luasteam.so
OS X AppIcon.icns

I assume the .love file is what the code needs to be adjusted to accept, but I'm no lua dev, so if this could be fixed that would be incredible.

erijohnt:
Did some cleanup to make this nicer. Tested on macOS but should work on Linux and Windows.

- Wrap the top-level in a `main()` function
- Remove dead commented code
- Get OS type (windows/macOS/linux) for upfront convenience
- Put 7zip download in its own function
- Put 7zip command in its own function
- Use context managers for tmp dirs (nice for auto-cleanup)
- Get/create mods dir and print it for the user

Should fix #23, and incorporates @gsmoraes's idea to auto-create and print the mods dir from [this comment](https://github.com/Steamopollys/Steamodded/issues/23#issuecomment-2033283714).

### DIFF
--- a/steamodded_injector.py
+++ b/steamodded_injector.py
@@ -1,11 +1,15 @@
+from pathlib import Path
 import subprocess
 import os
 import sys
-import tempfile
+from tempfile import TemporaryDirectory
+from typing import Literal
 import zipfile
 import shutil
 import platform
+from contextlib import contextmanager
 import requests
+
 
 def download_file(url, output_path):
     response = requests.get(url, stream=True)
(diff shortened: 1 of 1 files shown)


erijohnt:
Oh looks like @gsmoreas figured out that 7zip isn't actually required at all on macOS? That makes things even simpler, but I'm too lazy to fix the updated script because I already got way distracted by this, lol.

Steamopollys:
Hey, since it's not an SFX on mac, it don't surprise me that it can be manipulated without 7zip.

I will review this when I can!

Aurelius7309:
The injector is no longer to be used, starting on the upcoming major release.

PR #72, opened 2024-04-07
Adds payout screen API + `calc_dollar_bonus` function for jokers

GauntletGames-2086:
Not really needed but thought "hey, if I can make a UI fix for the payout screen, why not just make API for it?"

Adds: 
- Payout_Args API. To simplify, this will let you add money and/or display messages through the payout screen. Example provided. 
- Adds a fix for the payout screen not properly displaying negative numbers. Before, losing money in the payout screen would result in the section being completely blank. `NegativeMoney_add_round_eval_row` is a fix to this. 
Adds `calc_dollar_bonus` function to `joker.lua`. If we're going to make it easy to add money in the payout screen, might as well let jokers do it too. 

Notes: There could be potential issues with this API, `r_blind.lua` overrides the `add_round_eval_row` function that this API has to use. Couldn't find any errors in my initial check, but there could be a problem with this. I also might've missed something with the API setup or features, never made whole API before. 

image

### DIFF
--- a/core/StackTracePlus.lua
+++ b/core/StackTracePlus.lua
@@ -443,25 +443,255 @@ Stack Traceback
     return _M
 end
 
-
 -- Note: The below code is not from the original StackTracePlus.lua
 local stackTraceAlreadyInjected = false
+
+function getDebugInfoForCrash()
+    local info = "Additional Context:\nBalatro Version: " .. VERSION .. "\nModded Version: " .. MODDED_VERSION
+    local major, minor, revision, codename = love.getVersion()
+    info = info .. string.format("\nLove2D Version: %d.%d.%d", major, minor, revision)
+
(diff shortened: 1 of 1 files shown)


GauntletGames-2086:
Forgot to add negative money compatibility to payout_args, whoopsie

GauntletGames-2086:
- Added `above_dot_bar` so messages can be positioned before the dot bar UI
- Made the `payout` function less scuffed

Aurelius7309:
Would you be up to update this once more? I can convert the API itself to the new format no problem, but I'd appreciate if the other changes made here were provided as lovely patches.

GauntletGames-2086:
https://github.com/Steamopollys/Steamodded/pull/88 Moved to this pull request due to different fork.

PR #73, opened 2024-04-07
7zip fix on macos and readme isntructions

ManveerBhullar:
Fix for macos and also readme to be clearer.


kir68k:
I created a PR on your `macos-7zz` branch/repo that adds this fix for Linux as well, which I've gotten and there's an issue on, and fixed incorrect markdown formatting in the README.

Preferring this over creating another PR here :)

PR #74, opened 2024-04-08
Include 7-zip to injector

jie65535:
This PR avoids repeatedly downloading 7z during each injection.


Steamopollys:
Hey, keeping 7zip out of the bundle was a choice, I don't want to include it in the binary.

Thanks for the PR, but I'm not planning to add 7zip directly in the bundle.

jie65535:
I respect your choice, and I will keep the fork synchronization release containing 7z on my own.

This is because in China, for some reasons, most people cannot access Github normally.

Steamopollys:
Maybe lovely is a better solution in this case no?

Steamopollys:
I think it, is: https://discord.com/channels/1116389027176787968/1217867725532041369

PR #75, opened 2024-04-09
Update Lovely Injector Instructions in README.md

DMonitor:


DMonitor:
The current instructions are a very unclear. Lovely seems to be the superior way of installing mods, so it having the worst instructions is a travesty.

Steamopollys:
Hello,

We will hold on that until the experimental version of Balatro become the official one.

PR #77, opened 2024-04-12
Add Sticker API

GauntletGames-2086:
Adds sticker API. 

SMODS.Sticker:new variables: 
- `name`: Sticker name
- `key`: The sticker's label. Used for assigning sprites and variables. 
- `config`: Used for when sticker effect needs to store vars within itself. 
- `pos`: SpritePos
- `chance`: The rate of this sticker showing up in the shop. This is seeded automatically
- `atlas`: Sprite atlas
- `color`: The color of the sticker badge
- `default_compat`: When jokers are loaded, they are automatically given a custom sticker_compat variable. This dictates if said variable is auto set to `true` or `false`
    - NOTE: custom sticker_compat when loaded is set to `(sticker.key).."_compat"` (i.e. `example_sticker_compat`)
- `compat_exceptions`: A dictionary that sets each entry's `sticker_compat` to that value instead of being set to `default_compat`. 

Functions: 
- `set_sticker(card, case)`: Logic for determining if the sticker is set and any immediate effects (such as setting the cost of the joker). 
    - To make the joker considered to have your sticker, you must set `card.ability.sticker_label` to either a boolean or `sticker.config`
- `calculate(card, context)`: Lets you calculate a sticker's effect. Note: DO NOT RETURN
- `tooltip(card, info_queue)`: Assigns the tooltip for the sticker. No tooltip, don't define this function
- `set_badge(card, badges)`: Assigns the sticker's badge to the joker. No sticker badge, don't define this function
- `set_shader`: Assigns custom shader to the sticker. Not defining this function sets the sticker shader to the default shader

Hopefully I didn't miss something super ultra grave (clueless)


Aurelius7309:
I haven't been able to test this fully yet, but a couple of things stood out to me scanning over your changes. 

* There's an inconsistency on what the sticker's calculate function needs to be called (`calculate_joker` needs to exist and be a function in order for `calculate_sticker` to be called).
* If I didn't miss some change you made to the main function, copying/pasting the code of `create_card` seems unnecessary and unstable towards future updates of the game, given you could just take the card returned from the original function and place the functionality that's actually added (i.e., setting the stickers) in a hook after that function call.
* Your API currently offers no way for jokers to provide their own compatibility settings for custom vouchers without messing with sticker data that may or may not exist depending on loading order. You don't have to worry about integrating it into the Joker API since that's getting changed to be more flexible concerning optional parameters anyways, but `injectStickers` definitely should happen after `injectJokers` to ensure jokers can override the default values.

Thanks for taking the time to create this, I'll have a closer look in testing to make sure everything else works as intended.

Aurelius7309:
One more consideration, shouldn't a sticker's calculate function also not be called when the joker is debuffed?

Aurelius7309:
The implementation of `auto_set_true` and `auto_set_false` also seems inefficient, given how you loop over these tables for every joker for every sticker. A different representation, such as a set (storing the slugs as keys of the table instead of values in an array), would allow for instant access to the needed information (whether a specific slug is contained). Furthermore it occurred to me that it's not actually correct to call injectStickers before injectJokers unless default values are also assigned in injectJokers. Instead, it needs to happen after, but not assign the default compat value if *any* value is already present. This means checking for nil instead of any falsy value

GauntletGames-2086:
> One more consideration, shouldn't a sticker's calculate function also not be called when the joker is debuffed?

Pretty much none of the stickers currently in the game are restricted by that. Rental will still lose you money regardless of if the joker is debuffed or not.

PR #93, opened 2024-05-07
Stake API

MathIsFun0:
Mostly complete, but there are a few known issues:
- Switching/creating profiles causes the API to completely break
- Decks and achievements that require beating certain stakes don't update when new stakes are added in-between existing ones.


MathIsFun0:
Merging this as all the major bugs have been fixed

PR #99, opened 2024-05-16
use display name for previously applied Stakes

MathIsFun0:


MathIsFun0:
this is probably easier to add if I commit it directly

PR #103, opened 2024-05-17
Implement Enhancements

english5040:
A lot of stuff would need to be done manually if this code were used as is.

Next
- rationalize existing Enhancements;
- support enhancements that override the base card vs ones that don't. this is relevant for tooltips too


english5040:
Something is up with the diff. probably a broken rebase somewhere

english5040:
The problem may be a change in line endings

PR #127, opened 2024-06-18
Create zh_CN.lua

ChromaPIE:
zh_CN (Simplified Chinese) loc


english5040 (approved):
Alright, since this is a language file I will just trust that the localization is good. Can merge

PR #129, opened 2024-06-22
Pools, fix for playing cards

english5040:


english5040:
Using in_pool when setting the initial deck could have weird behavior, since technically the game hasn't started. We can either set some flag in G.GAME that indicates whether the game has started, or have an in_initial_deck flag

PR #134, opened 2024-06-24
Preliminary support for localizing Palette

HuyTheKiller:
I need some help to further expand the usability of these localizing variables. Currently "Card Colours" can be localized, while anything inside are broken by now.


HuyTheKiller:
I've decided to revert the changes to anything lying under the Card Colours button, only keeping the variables and the patch.

PR #135, opened 2024-06-25
feat: prevent crash with corrupt/invalid saves

WilsontheWolf:
This pr adds a check to the game that prevents it from crashing when pressing the play button with an invalid/corrupted save.jkr, allowing people to make a new game. This is technically a balatro bug, but invalid saves are more common when using mods, so I figured it would be nice to fix.

A before and after:

https://github.com/Steamopollys/Steamodded/assets/33164598/db323baf-b68b-47c5-95bf-3c1210c6d1ae

Ping me on discord if you want a corrupt save to test with (github doesn't let me upload .jkr files)


english5040 (approved):
Ok, this looks fine to merge

PR #138, opened 2024-06-27
Fix shader keys auto adding mod prefix

Eremel:
Shader keys should now be referenced as `modprefix_shaderkey`

### DIFF
--- a/core/game_object.lua
+++ b/core/game_object.lua
@@ -22,7 +22,7 @@ function loadAPIs()
     function SMODS.GameObject:__call(o)
         o = o or {}
         o.mod = SMODS.current_mod
-        if o.mod and not o.raw_atlas_key and not (o.mod.omit_mod_prefix or o.omit_mod_prefix) then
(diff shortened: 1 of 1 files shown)


core/game_object.lua
```diff
@@ -22,14 +22,11 @@ function loadAPIs()
     function SMODS.GameObject:__call(o)
         o = o or {}
         o.mod = SMODS.current_mod
-        if o.mod and not o.raw_atlas_key and not (o.mod.omit_mod_prefix or o.omit_mod_prefix) then
+        if o.mod and not o.raw_atlas_key and not (o.mod.omit_mod_prefix or o.omit_mod_prefix) and not o['shader'] then
```
Aurelius7309, line 25: Why should a `shader` being present prevent atlas keys from being prefixed?

Eremel: If the object is a shader, the atlas refers to how it appears in the collection only. Removing the mod prefix in that situation lets you use whichever atlas you like to display the shader, otherwise it’s assigned the mod prefix automatically

PR #139, opened 2024-06-28
allow blinds' loc_txt to have more than 2 lines

Mysthaps:
need a bit more testing likely


english5040:
This PR still needs some help injecting into locations where [1] and [2] are used directly

Aurelius7309 (approved):

PR #149, opened 2024-07-01
Add keybinds API

stupxd:
This will add ability to create keybinds. For now keybinds will only run on key press, but it's possible to add support for key release and hold (not sure how necessary this is, so didn't feel like adding it yet).

Includes an example mod to showcase the new functionality.


english5040:
Seems simple enough to put in directly. thanks

I'll squash merge

PR #151, opened 2024-07-01
Alt texture new

Eremel:


core/game_object.lua
```diff
@@ -2419,165 +2423,237 @@ function loadAPIs()
...
     -------------------------------------------------------------------------------------------------
 
-    SMODS.local_palettes = {}
-    SMODS.Palettes = { Types = {} }
-    SMODS.Palette = SMODS.GameObject:extend {
-        obj_table = SMODS.local_palettes,
+    SMODS.alt_textures = {}
+
+    SMODS.AltTextures = {
+        Types = {"Suit", "Tarot", "Planet", "Spectral", "Joker", "Enhanced", "Back", "Seal", "Voucher", "Booster", "Tag", "Blind"},
```
english5040, line 2432: The organization here is a bit weird, Types is not a 'real' element of the table

core/game_object.lua
```diff
@@ -2419,165 +2423,237 @@ function loadAPIs()
...
+        Joker = {names = {}}, Voucher = {names = {}}, Booster = {names = {}}, Enhanced = {names = {}},
+        Back = {names = {}}, Tag = {names = {}}, Seal = {names = {}}, Blind = {names = {}}
+    }
+
+        
+    ---@param key string @Missing required parameter for %s declaration: %s
+    ---@param type string
+    ---@param name string
+    SMODS.AltTexture = SMODS.GameObject:extend {
+        obj_table = SMODS.alt_textures,
```
english5040, line 2443: Just use {} here, if the SMODS.alt_textures name is not intended to be used

core/game_object.lua
```diff
@@ -2419,165 +2423,237 @@ function loadAPIs()
...
+            -- Initialize the new palette
+            if self.palette then self:create_palette()
+            elseif self.texture then self:create_new_atlas() end
         end
     }
 
-    function SMODS.Palette:create_default(type)
-        table.insert(SMODS.Palettes[type].names, "Default")
-        SMODS.Palettes[type]["Default"] = {
+    function SMODS.AltTexture:create_palette()
```
english5040, line 2483: from_palette, from_atlas seem more natural. after all, we aren't creating a palette, we're creating a texture from a palette

core/overrides.lua
```diff
@@ -1017,87 +1016,87 @@ end
...
+                type = 'title',
+                highlight_limit = 0,
+                collection = true
+            })
+        table.insert(deck_tables, { n = G.UIT.R, config = { align = "cm", padding = 0, no_fill = true },
+            nodes = {{ n = G.UIT.O, config = { object = G.your_collection[j] } }}
+        })
+    end
+
+	table.sort(G.P_CENTER_POOLS.Enhanced, function(a,b) return a.order < b.order end)
```
english5040, line 1037: Sort Enhancements when they're initially defined, not here

core/utils.lua
```diff
@@ -388,73 +388,93 @@ function serialize_string(s)
...
 
+-- Called from option selectors that control each type
 G.FUNCS.update_recolor = function(args)
-    G.SETTINGS.selected_colours[args.cycle_config.type] = SMODS.Palettes[args.cycle_config.type][args.to_val]
+    G.SETTINGS.selected_texture[args.cycle_config.type] = args.to_val
 	G:save_settings()
 	G.FUNCS.update_atlas(args.cycle_config.type)
 end
 
 -- Set the atlases of all cards of the correct type to be the new palette
```
english5040, line 422: slight error in comment: palette -> 'texture'

core/utils.lua
```diff
@@ -388,73 +388,93 @@ function serialize_string(s)
 	return string.format("%q", s)
 end
 
---#region palettes
-G.SETTINGS.selected_colours = G.SETTINGS.selected_colours or {}
+--#region alt textures
+G.SETTINGS.selected_texture = G.SETTINGS.selected_texture or {}
 G.PALETTE = {}
```
english5040, line 393: is this needed?

core/utils.lua
```diff
@@ -388,73 +388,93 @@ function serialize_string(s)
...
-		local v = SMODS.Palettes[k]
-        if #v.names > 1 then
-            nodeRet[#nodeRet+1] = create_option_cycle({w = 4,scale = 0.8, label = k.." colours" ,options = v.names, opt_callback = "update_recolor", current_option = G.SETTINGS.selected_colours[k].order, type=k})
-        end
-    end
-    local t = create_UIBox_generic_options({back_func = 'options', contents = nodeRet})
-    return t
-end
-
 G.FUNCS.recolour_image = function(x,y,r,g,b,a)
```
english5040, line 461: Not sure what this is doing now

core/game_object.lua
```diff
@@ -805,22 +809,16 @@ function loadAPIs()
             SMODS.process_loc_text(G.localization.descriptions.Other, 'undiscovered_' .. string.lower(self.key),
                 self.loc_txt, 'undiscovered')
         end,
-        generate_colours = function(self, base_colour, alternate_colour)
-            if not self.colour_shifter then return HEX("000000") end
+        generate_colours = function(self, base_colours)
```
english5040, line 812: Concerned about adding this- I don't want to add things to SMODS that are not generally useful utilities, this seems pretty specific.

core/game_object.lua
```diff
@@ -2302,10 +2310,6 @@ function loadAPIs()
             SMODS.process_loc_text(G.localization.misc.labels, self.key:sub(3), self.loc_txt, 'label')
             SMODS.Edition.super.process_loc_text(self)
         end,
-        -- apply_modifier = true when G.GAME.edition_rate is to be applied
-        get_weight = function(self, apply_modifier)
```
english5040, line 2306: Why was get_weight removed? also get_weight remains a bad name, sorry. to be honest, letting weight itself be either a number or a function might be better

core/game_object.lua
```diff
@@ -2419,165 +2423,237 @@ function loadAPIs()
...
+        else
+            atlas_key = self.type
         end
-        return { HEX(base_colour) }
+        G.ASSET_ATLAS[atlas_key][self.name] = G.ASSET_ATLAS[self.atlas_key]
     end
 
-    for k, v in pairs(G.P_CENTER_POOLS.Tarot) do
-        SMODS.Consumable:take_ownership(v.key, { atlas = "Tarot" })
+    -- Call the `generate_colours` function of the appropriate type
```
english5040, line 2570: Again, create_colours is something I am wary of including in Steamodded itself. I'd like these functions to be cohesive, any specific functions can go in a separate mod.

core/game_object.lua
```diff
@@ -2419,165 +2423,237 @@ function loadAPIs()
...
+        suits = {
+            Clubs = "008ee6",
+            Spades = "3c4368",
+            Diamonds = "e29000",
+            Hearts = "f83b2f"
+        },
+        type = "Suit",
+        name = "High Contrast",
+    })
+    for k,v in pairs(G.P_CENTER_POOLS.Tarot) do
```
Aurelius7309, line 2606: These `take_ownership` calls flood logs at startup for no good reason. Perhaps they should be made silent?

core/overrides.lua
Eremel: I haven't added this specifically, but I'm pretty sure it's here to make sure the Enhancements are in the correct order before displaying in the collection

core/game_object.lua
Eremel: I have no idea how this has been removed. I've renamed to `modified_weight` instead, but you need to be able to access the base weight and the modified weight to let `poll_edition` function correctly.

core/utils.lua
Eremel: this is the function that recolours base textures

core/utils.lua
Eremel: The saved settings don't exist without this and the game won't load, unsure if there's a cleaner way around this

core/overrides.lua
english5040: Ok, I guess we'll just add a TODO comment

core/game_object.lua
```diff
@@ -102,16 +102,12 @@ function loadAPIs()
...
-                ('Registered game object %s of type %s')
-                :format(o.key, o.set), o.set or 'GameObject'
-            )
+            if not o.no_log then sendInfoMessage(('Registered game object %s of type %s'):format(o.key, o.set), o.set or 'GameObject') end
         end
     end
 
     --- Takes control of vanilla objects. Child class must implement get_obj for this to function.
-    function SMODS.GameObject:take_ownership(key, obj, silent)
+    function SMODS.GameObject:take_ownership(key, obj, silent, no_log)
```
english5040, line 110: for the no-logging functionality, I was thinking of a setting on the logging system, not a setting per-object. basically, some toggle that temporarily prevents logging. It could be as simple as a global variable- that's not extensible, but it's also simple enough to be fine

Issue #144, opened 2024-07-01
Rationalize prefixes


english5040:
discussed- have add_prefix_to_key as a common function for each class, with a default implementation; have 'class' and 'mod' prefixes.

Prefixes to add can be specified per class as an unordered table

PR #154, opened 2024-07-02
Rationalize prefixes

Aurelius7309:
Allows for consistent configuration of key prefixing across all classes (#144).
Changes `card_key` for suits and ranks to comply with this. This field is no longer subject to automatic decollisioning and uses mod prefixes exclusively for this purpose.
TODO:
- ~~`SMODS.GameObject:take_ownership()` is currently a huge mess. We need to find a better way to integrate it into this new system.~~

### DIFF
--- a/core/core.lua
+++ b/core/core.lua
@@ -5,7 +5,7 @@ SMODS = {}
 SMODS.GUI = {}
 SMODS.GUI.DynamicUIManager = {}
 
-MODDED_VERSION = "1.0.0-ALPHA-0702a-STEAMODDED"
+MODDED_VERSION = "1.0.0-ALPHA-0706a-STEAMODDED"
 
 function STR_UNPACK(str)
 	local chunk, err = loadstring(str)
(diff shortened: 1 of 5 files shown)


Aurelius7309 (commented):
This currently crashes because it breaks something in Palettes

english5040:
Palettes can be removed if it breaks and isn't easily fixable; AltTexture is going to replace it.

(still would be good for merge conflict reasons if palettes can be fixed)

Aurelius7309 (commented):
The issue was in fact not with Palettes, but because default atlases were incorrectly being prefixed.

Aurelius7309:
Stakes are fixed. Taking ownership of the data in place causes the stake level to come out wrong, causing other crashes, so this approach was not taken.

Aurelius7309:
There's a different issue relating to take_ownership where Blinds that have been taken ownership of are automatically discovered: #157. This doesn't happen to any other object types.

english5040:
The fact that taking ownership of stakes is broken is definitely an issue to be fixed

english5040 (approved):
PR is in a good state

PR #169, opened 2024-07-09
SMODS.Seal - Better calculate args, allow generic modifiers (mult, chips, etc.) for seals, example seal mod

stupxd:
This should make the seals `calculate` function consistent with other game object APIs.

Also I've spent a lot of time trying to make seals work, as I couldn't find any proper example for 1.0.0, hopefully this example mod is helpful to someone.

Also ignore 3 commits for keybinds API, no idea why git thinks they're new, as they're merged already


english5040:
Got some whitespace changes in the PR and your seal update may clobber a recent change to seals, I'll clean those up for you.

I'm concerned that returning values in seal calculation will start a pattern of people depending on that behavior. Better calculation can come later though, I see it's consistent with other base game things

stupxd:
> your seal update may clobber a recent change to seals
Not sure which one do you mean. The one that makes seals calculate function to be (self, card, context) instead of (card, context)?

english5040:
Thank you for the PR

PR #172, opened 2024-07-10
Replace FiveOfAFlush example with RoyalFlush

MathIsFun0:
The original Five of a Flush mod I created dates back to the September demo. Since that point, the mod has become completely incompatible and Flush Five was added as a hand type.

This pull request replaces this mod with a mod that works in the latest version of Steamodded 1.0.0-alpha and adds the Royal Flush as its own separate poker hand rather than a cosmetic version of the Straight Flush.


english5040:
Alright, thanks for the PR. This looks simple enough to merge right away

PR #173, opened 2024-07-10
SMODS.Enhancement - Add process_loc_text for tooltips

stupxd:
Noticed that this was missing, seems like an oversight


english5040:
yeah, this is probably because enhancements were jank before and got ignored.

I don't believe the :lower is needed; also why can't tooltips use the vanilla method of using the center as an element in info_queue?

english5040:
I don't believe this is needed

PR #176, opened 2024-07-11
SMODS.Enhancement - Fix crash when using enhancement prototype for tooltips

stupxd:


english5040:
ok to merge, as a temporary measure

PR #180, opened 2024-07-13
Example Editions mod improvements

stupxd:
Better shaders explanation
remove unnecessary functions (mostly RGB and HSL, where they're not used)
fix flipped shader (it never worked)
improve readme


english5040:
alright, thanks

PR #181, opened 2024-07-13
Mod badge height fix

Eremel:
Stops mod badges from being scaled down in height when the name is long.
Added `TEXT_COLOUR` argument to mod headers to set the mod badge text colour.

*ignore the shader crap my git is screwed up (:*


english5040:
Bad name for TEXT_COLOUR, since it might easily conflict. Since it's about badges, changing to BADGE_TEXT_COLOUR

PR #186, opened 2024-07-15
feat: SMODS.load_file

WilsontheWolf:
Add's a new method, `SMODS.load_file(path, id)`. This method take's in a path, relative to the mod's main directory, and optinally the mod id (will use SMODS.current_mod if not. Just got thinking should I require the mod object instead of the id?). It then reads the file and loads it. Return type is the same as [love.filesystem.load](https://love2d.org/wiki/love.filesystem.load) and [loadfile](https://www.lua.org/manual/5.1/manual.html#pdf-loadfile).

Example:
`Main.lua`
```lua
--- STEAMODDED HEADER
--- MOD_NAME: Load Test
--- MOD_ID: LoadTest
--- MOD_AUTHOR: [WilsontheWolf]
--- MOD_DESCRIPTION: Testing for new api
----------------------------------------------
------------MOD CODE -------------------------
local mod = SMODS.current_mod;

assert(SMODS.load_file("Loaded.lua"))()
----------------------------------------------
------------MOD CODE END----------------------
```
`Loaded.lua`
```lua
print("Hi mom")

-- error("Test") -- Uncomment me to see that my filename is properly shown in the crash log
```

- Also fixes a parsing issue in StackTracePlus, which I encountered testing if this worked.
- Also nil's current_mod after loading mods, which prevents mods from accidentally loading files from another mod's directory.

### DIFF
--- a/core/StackTracePlus.lua
+++ b/core/StackTracePlus.lua
@@ -353,12 +353,27 @@ Stack Traceback
                 if string_sub(info.source, 1, 1) == "@" then
                     dumper:add_f("(%d) main chunk of file '%s' at line %d\r\n", level_to_show,
                         string_sub(info.source, 2), info.currentline)
-                elseif info.source and info.source:sub(1, 21) == "--- STEAMODDED HEADER" then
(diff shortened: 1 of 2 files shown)


english5040:
Merged

PR #203, opened 2024-07-21
Nuke lua reloading entirely from orbit

flakywanderer:
As mentioned to me, this functionality is broken and already on its way out, and it's currently causing incompatibilities, such as with [https://github.com/OceanRamen/BU-CB](https://github.com/OceanRamen/BU-CB). 

I think it's about time to remove it.


english5040:
Yeah this is just causing problems, merging

PR #204, opened 2024-07-21
Revise file loading / structure

Aurelius7309:
- Move most code from `core.lua` to `ui.lua`
- `core.lua` is now responsible for loading lovely/nativefs libs, setting mod dir and loading the other files
- Use NFS instead of lovely patches to load other files to improve crash readability
  - Error messages: `main.lua:3898: Missing required parameter ...` -> `[SMODS _ "core/game_object.lua"]:28: Missing required parameter ...`
  - Stack trace: `[6] Lua global 'loadMods' at Steamodded file 'loader/loader.lua:264'`
 - Disallow library load failure as nativefs now comes with Steamodded


english5040:
I would take a look at the code setting NFS's current directory. Mods may have unexpected behavior if they're using NFS as a replacement for love.filesystem otherwise

loader/loader.lua
```diff
@@ -1,83 +1,6 @@
...
-end
-if nfs_success then
-    NFS = nativefs
-    -- make lovely_mod_dir an absolute path.
-    -- respects symlink/.. combos
-    NFS.setWorkingDirectory(lovely_mod_dir)
-    lovely_mod_dir = NFS.getWorkingDirectory()
-    -- make sure NFS behaves the same as love.filesystem
-    NFS.setWorkingDirectory(love.filesystem.getSaveDirectory())
-else
```
english5040, line 24: Concerned about removing the block above.

Aurelius7309: The relevant code wasn't removed, but relocated to `core.lua` because it needs to run earlier

PR #207, opened 2024-07-23
Fix SMODS.Sticker + Stickers Collections Tab

GauntletGames-2086:
Changes: 
- Fixes the SMODS.Sticker API. 
- Adds Stickers collections tab to the "Other" objects button.
- Better disabling of the Additions tab (doesn't use mod.added_obj now, also addresses a bug relating to only Red Deck when opening "New Run"). 

TODO: 
- ~~Create stickers for all the currently existing ones (Eternal, Persiahable, Rental, Pinned) so others can take ownership of them.~~ 
- ~~Make the `create_card` check a function within the SMODS.Sticker API instead of separate (this is to allow more complex behavior when determining whether a card has that sticker or not, such as "All X Sticker" support).~~


core/ui.lua
```diff
@@ -323,7 +324,7 @@ function buildAdditionsTab(mod)
 
 	local modNodes = {}
 	table.insert(modNodes, t)
-	return {
+	return (#leftside_nodes > 0 and #rightside_nodes > 0 ) and {
```
Aurelius7309, line 327: This doesn't seem right...? It's possible for one column to be empty and the tab should still show

GauntletGames-2086: It isn't, this should be `or` instead of `and`.

GauntletGames-2086:
Whether the 

> Be careful when fixing the enhancements UI bug- it should "just work" in the same way as I fixed seals a while ago. A hacky solution is ok for now but if it's buried in other code, either move it elsewhere or leave it out of the PR
> 
> I'll take a look at enhancements

I'll move it out; the solution itself edits too much for a PR directed at fixing SMODS.Sticker spaghetti. I do think that for it's generate_ui function it should NOT be referring back to SMODS.Center's generate_ui function. This UI bug is already caused due to this (the game is generating full_UI_table.name in collections when it should not) and parts of SMODS.Center itself is only related to SMODS.Enhancement UI.

core/game_object.lua
```diff
@@ -2148,9 +2146,85 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
...
+    }
+
+    -- Create base game 
+    -- sticker_check is set to nil on all since they'd use base game checks for application
+    SMODS.Sticker{
+        key = "eternal",
+        badge_colour = HEX 'c75985',
+        prefix_config = {key = false},
+        pos = { x = 0, y = 0 },
+        sticker_check = nil,
```
Aurelius7309, line 2180: `sticker_check = nil` does nothing, it acts like there is no value there. Use 0 or false instead. Though the base checks should work here?

core/ui.lua
```diff
@@ -494,6 +527,157 @@ G.UIDEF.consumable_collection_page = function(page)
...
+			local idx = i + (j - 1) * cols + offset
+			if idx > #sticker_pool then return end
+			local center = sticker_pool[idx]
+			local card = Card(G.your_collection[j].T.x + G.your_collection[j].T.w / 2, G.your_collection[j].T.y,
+				G.CARD_W, G.CARD_H, G.P_CARDS.empty, G.P_CENTERS["c_base"])
+			card.ignore_pinned = true -- Scuffed solution to ignoring the effect of pinned, I'll figure out something better later
+			center:set_sticker(card, true)
+			G.your_collection[j]:emplace(card)
+		end
+	end
```
Aurelius7309, line 679: Missing 'end' here

core/game_object.lua
GauntletGames-2086: Thunk already has sticker checks for these 3 and does some funkyness with the sticker check for eternal/perishable (they use the same pseudorandom var), so sticker_check function is removed as to preserve base game sticker application. I'll just remake the sticker check function instead (assuming eternal/perishable funkyness doesn't stop that).

GauntletGames-2086: Update: psuedorandom does change when calling the function again therefore adding `sticker_check` as a function would mean different RNG for their application. Not sure on the opinion for this, so defaulting to `sticker_check = false` for both.

Aurelius7309:
If there's still some issues with this, I've missed them and will fix myself when/if they show up. Looks good for the moment, merging

Aurelius7309 (approved):

PR #211, opened 2024-07-24
Convert sound API to using sound_manager

Aurelius7309:
Sound API currently has issues playing music and bypasses the modulate functionality completely. This integrates it with the sound manager at hand.

### DIFF
--- a/core/game_object.lua
+++ b/core/game_object.lua
@@ -354,10 +354,6 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
                 sendWarnMessage(('Detected duplicate register call on object %s'):format(self.key), self.set)
                 return
             end
-            if self.language then
-                -- TODO localized sounds are wonky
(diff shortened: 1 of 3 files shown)


MathIsFun0:
Trying to play a custom sound and getting an error on this line:
`if string.find(s.sound_code,'music') then`
because `s.sound_code` is nil
https://github.com/MathIsFun0/Cryptid/tree/sound-rework (I was using Big Cube to test if even the XChip sound would work)

Aurelius7309:
Oops, fixed

MathIsFun0:
Can confirm it works, a few things I want now though:
- an option to apply the pitch shifting or not
- an option to force sync with the main music tracks (this could help music tracks remain synced for a full game session if they're a few frames shorter or longer than the vanilla tracks)

Aurelius7309:
- You can add `pitch` and `volume` fields to override the default music values, though I should probably make those work as defaults for other sounds
- It should sync the music when no_sync is absent regardless of length

Issue #155, opened 2024-07-03
Weights on editions, packs, etc. should be consistently named and used


english5040:
We are starting to get more things with weights. We should come up with some consistent system for weighted selection

PR #212, opened 2024-07-24
Seal weights and sounds

Eremel:
Adds `poll_seal` instead of hardcoded values or `rng_buffer`. Allows seals to have a `weight` value, and also respects `in_pool` results if called without a set of options.

Added custom sounds to seals when they are applied. Defaults to vanilla sound.

### DIFF
--- a/core/utils.lua
+++ b/core/utils.lua
@@ -584,3 +584,43 @@ function format_ui_value(value)
 end
 
 --#endregion
+
+
+function poll_seal(_key, _mod, _guaranteed, _options, _type_key)
+    _mod = _mod or 1
+    local seal_poll = pseudorandom(pseudoseed(_key or 'stdseal'..G.GAME.round_resets.ante))
+    -- If no _options provided, get all seal keys
+    if not _options then
+        _options = get_current_pool("Seal")
+	end
+
+    local available_seals = {}
+    local total_weight = 0
+    for _, v in ipairs(_options) do
+        if v ~= "UNAVAILABLE" then
+            local seal_option = {}
+            if type(v) == 'string' then
+                assert(G.P_SEALS[v])
+                seal_option = { name = v, weight = G.P_SEALS[v].weight or 25 }
+            elseif type(v) == 'table' then
+                assert(G.P_SEALS[v.name])
+                seal_option = { name = v.name, weight = v.weight }
+            end
+            if seal_option.weight > 0 then
+                table.insert(available_seals, seal_option)
+                total_weight = total_weight + seal_option.weight
+            end
+        end
+	end
+
+    if _guaranteed or seal_poll > 1 - (0.02*_mod) then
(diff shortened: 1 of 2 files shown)


core/utils.lua
```diff
@@ -584,3 +584,43 @@ function format_ui_value(value)
 end
 
 --#endregion
+
+
+function poll_seal(_key, _mod, _guaranteed, _options, _type_key)
```
Aurelius7309, line 589: Consistency, this should be `SMODS.poll_seal`. API design, this has too many arguments. I suggest converting to a table argument that can hold these arguments as needed

core/utils.lua
```diff
@@ -584,3 +584,43 @@ function format_ui_value(value)
...
+	end
+
+    local available_seals = {}
+    local total_weight = 0
+    for _, v in ipairs(_options) do
+        if v ~= "UNAVAILABLE" then
+            local seal_option = {}
+            if type(v) == 'string' then
+                assert(G.P_SEALS[v])
+                seal_option = { name = v, weight = G.P_SEALS[v].weight or 25 }
```
Aurelius7309, line 604: In reference to #155, this should expose the same `get_weight` function that editions do.

lovely/seal.toml
```diff
@@ -9,15 +9,35 @@ priority = 0
...
 payload = '''
-card:set_seal(SMODS.Seal.rng_buffer[math.ceil(seal_type*#SMODS.Seal.rng_buffer) or 1])'''
+card:set_seal(poll_seal(nil, 10))'''
+
+# Card:calculate_joker()
+[[patches]]
+[patches.regex]
+target = 'card.lua'
+pattern = '''
+(?<indent>[\t ]*)local seal_type = pseudorandom\(pseudoseed\('certsl'\)\)
```
Aurelius7309, line 32: Consistency: why does this exclude modded seals?

lovely/seal.toml
```diff
@@ -179,4 +199,32 @@ position = "after"
...
+
+# Game:init_item_prototypes()
+[[patches]]
+[patches.pattern]
+target = 'game.lua'
+pattern = '''table.sort(self.P_CENTER_POOLS["Seal"], function (a, b) return a.order < b.order end)'''
+position = 'after'
+match_indent = true
+payload = '''
+table.insert(G.P_CENTER_POOLS.Seal, 4, G.P_CENTER_POOLS.Seal[1])
```
Aurelius7309, line 212: Invalidates `order`, can we change the order on P_SEALS directly?

Aurelius7309 (changes requested):

lovely/seal.toml
Eremel: I was unsure whether or not to change this, do we want it to respect modded seals or act as vanilla behaviour in all circumstances?

Aurelius7309 (approved):
Looks good now, merging

PR #214, opened 2024-07-29
make negative edition more general by checking card.ability.cardarea parameter and card.area. Since cardarea isn't created out of a run, I set card.ability.cardarea to be a string and SMOD looks up G[card.ability.cardarea]. Also buy_from_shop moves consumeables to ability.cardarea if it exists.

betmma:


betmma:
oops the title is too long :(
G[card.ability.cardarea] is not very ideal but is the best I can come up with, so any other implementations are welcome

Aurelius7309:
I think a better solution would be to have a function defined on the center object retrieve the card area. This would also be easier to set up for all objects of a class. Though I'm also not entirely sure what the whole `card.ability.cardarea` is needed for in determining which card area to change, shouldn't it just always be the area the card is in?

Aurelius7309:
That title was a description 💀

betmma:
> what the whole `card.ability.cardarea` is needed for in determining which card area to change, shouldn't it just always be the area the card is in?
add_do_deck is called before cardarea:emplace(card) so in add_to_deck function card.area hasn't been set. I think another solution could be moving code that changes area size to cardarea:emplace

Aurelius7309:
I don't think that'll work well with debuffed cards

DigitalDetective47:
> shouldn't it just always be the area the card is in?

natural negatives in the shop shouldn't increase shop joker slots

PR #215, opened 2024-07-29
SMODS.Achievements

GauntletGames-2086:
Explained in name. Created inititally for Cryptid but porting here since other mods should be able to use it. At this point it's basically feature complete. 

Args: 
- `unlock_condition`: A function that returns true when correct conditions are met to unlock. 
- `hidden_pos`: Sprite pos for when the achievement is not earned and displayed in the achievements tab. 
- `bypass_all_unlocked`: Ignore checks preventing achievement from being earned on profiles that used the "Unlock All" button. 
- `hidden_name/hidden_text`: Hides the name/desc of the achievement when not earned displayed in the achievements tab. 
- `reset_on_startup`: Unearns the achievement when injected.


GauntletGames-2086:
API is basically done! All that's left that I know of is just one visual bug with the tab (if name is longer than the wrapped desc the name UIBox extends further than the desc UIBox but not the other way around).

Aurelius7309 (approved):
Made some changes to the UI to have things looking clean, looks good otherwise. Merging

PR #217, opened 2024-07-29
Rationalize poker hands

Aurelius7309:
- Register vanilla poker hands as `SMODS.PokerHand` objects and rewrite `evaluate_poker_hand` accordingly
- New object type `SMODS.PokerHandPart`. This is simply a function container, all functions registered here are evaluated as parts available to access during evaluation of all poker hands.
- Evaluation of all poker hands is unified to one function `evaluate(parts, hand)`. Usage of parts is recommended, but the full hand is still provided as to not limit functionality. **This is a breaking change! Existing poker hands with `atomic_part` or `composite` functions will crash the game on startup!**
- Add utility method `SMODS.merge_lists` to merge numerically indexed tables in order without duplicates. This is more efficient than the vanilla algorithm for merging straight flushes, and should also be used when it would normally be expected one hand type should fully contain the other. For example, this allows fully scoring Flush Fives with 6 or more cards where the Flush and the 5oak only overlap partially
- Remove any hard-coded contained hands by having get_X_same return higher-order pairs than requested
- Two Pair and Full House share an `_all_pairs` part that merges all pairs of 2 or more cards, distinguish them by means of presence of a 3oak. This implementation is more general and does not fail to count higher amounts of pairs as Two Pair.
- Remove the `h_` prefix to keep consistency with vanilla hands


core/game_object.lua
```diff
@@ -1939,40 +1939,57 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
...
             'example',
+            'evaluate'
         },
         order_lookup = {},
+        max_order = { value = 0 },
         visible = true,
         played = 0,
         played_this_round = 0,
         level = 1,
-        class_prefix = 'h',
```
english5040, line 1960: I support this, since it was confusing. Reminder that we should warn people about this

core/game_object.lua
```diff
@@ -1939,40 +1939,57 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
...
+                    for i, v in ipairs(G.handlist) do
+                        if v == self.above_hand then j = i end
+                    end
+                    table.insert(G.handlist, j, self.key)
+                    self.order_lookup[self.above_hand] = (self.order_lookup[self.above_hand] or 0) - 0.001
+                    self.order = self.obj_table[self.above_hand].order + self.order_lookup[self.above_hand]
+                else
+                    self.max_order.value = self.max_order.value + 1
+                    self.order = self.max_order.value
+                    G.handlist[#G.handlist+1] = self.key
```
english5040, line 1991: An idea I had for poker hands: auto-sort them based on the product of base mult/chips; above_hand (if set) should be considered first. If two hands are both set to be above the same hand, there could also be conflicts?

core/game_object.lua
```diff
@@ -1986,6 +2003,70 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
...
+        key = '_straight',
+        func = function(hand) return get_straight(hand) end
+    }
+    SMODS.PokerHandPart {
+        key = '_flush',
+        func = function(hand) return get_flush(hand) end,
+    }
+    -- all sets of 2 or more cards of same rank
+    SMODS.PokerHandPart {
+        key = '_fh',
```
english5040, line 2020: A bit misleading to call this _fh if it's not necessarily a full house

Edit: yeah, if it's called _fh people might use it when it's just a pair. something like all_pairs or something might be better

english5040:
For a full overhaul, we'll want the part functions to behave correctly: overlapping flushes/2 disjoint straights should be handled correctly, with all possibilities going into _straight/_flush.

We should consider how parts work: do we ask that people merge all possible flushes, similar to merging _2s, to get the list of scored cards?

core/game_object.lua
Aurelius7309: Renamed it to `_all_pairs`

core/game_object.lua
Aurelius7309: I went ahead and added this, should also be more efficient than inserting each hand individually where it belongs. `order_offset` can be used to change ordering as needed. Multiple hands with the same above_hand are sorted the same way with lower priority

Aurelius7309:
> For a full overhaul, we'll want the part functions to behave correctly: overlapping flushes/2 disjoint straights should be handled correctly, with all possibilities going into _straight/_flush.
> 
> We should consider how parts work: do we ask that people merge all possible flushes, similar to merging _2s, to get the list of scored cards?

I went ahead and changed the implementation of merging parts to account for this, but notably straights and flushes will never return multiples. That shouldn't be too bad to add for flushes though, and straights need to be revisited anyways.

english5040:
(yeah, straights can come later)

PR #218, opened 2024-07-29
Earlier assignment of seed-generation in Game:start_run

larswijn:
Currently, `Game:start_run` first applies any stake and deck modifiers, and only assigns the seed afterwards. Any randomness used for any of these modifiers cannot make use of the seed of that run.

By replacing the original seed generation with an empty line, and adding a hard-coded copy earlier in the function, the seed is assigned earlier and can be used by both stakes and decks when applying to the start of the run.


Aurelius7309:
Note: playing cards are still generated after apply_to_run, and Game:start_run() relies on this ordering. There needs to be a better way for decks to modify the starting deck.

Aurelius7309:
Deck manipulation already requires using events, and the game's seed is available when the event fires. Thus this is not needed at this time, but a better way to implement deck effects will come later

PR #224, opened 2024-08-07
Retrigger API

MathIsFun0:
Allows for retriggering Jokers. In the process, also cleans up calculation a bit so a callback function is used for certain effects.

Needs more testing, but considering the demand for mods besides Cryptid to have this as a mechanic, it feels necessary to have this in Steamodded.


MathIsFun0:
"calculate_joker_retriggers" should probably be changed to instead be a way for editions to add retriggers, similar to how "calculate_blurred" just got changed in Cryptid's code

PR #225, opened 2024-08-07
feat: don't show empty additions tab

WilsontheWolf:
This is just a simple check that doesn't add the additions tab if both columns are empty. Realistically, the `added_obj` check should still be improved, but this handles the case where it's wrong.


Aurelius7309:
Addressed by #207, which is currently under review

PR #237, opened 2024-08-13
feat: crash handler better lovely mods

WilsontheWolf:
This improves the logic for handling lovely mods in the crash handler. Most notably, lovely mods now show up when the game crashes while loading mods, and lovely mods now show up for SMODS mods that couldn't load (missing dependencies for example).


Aurelius7309:
Ideally we don't want disabled mods to show up on the crash screen, but they are also marked as not being able to load. Good to go if this case is handled

PR #238, opened 2024-08-13
Add injection for eternal/perishable/rental globals

rileyinman:
Using a simple mod that replaces stickers with smaller versions (Less Intrusive Stickers by FabianPEKS) by replacing the stickers.png Atlas, non-stake stickers show as doubled:

image

This change fixes the issue by injecting the respective global variables in the Sticker object. This was the quickest and most obvious way I could see to fix the problem, but my experience with Lua and this game are low so let me know if you have requested changes. Thanks!


Aurelius7309:
Made some changes to bring this more in line with inject functions being either universal or customized for an object. Looks good otherwise, merging

rileyinman:
Thank you!

PR #254, opened 2024-09-04
SMODS.Shader - optional method for custom external vars

stupxd:
`send_vars = function (sprite)` - support custom external vars to be sent to shader. 
If a table is returned, `Shader:send(key, value)` will be run for every table entry.

Usage example:

`mod.lua`
```
SMODS.Shader {
    key = 'chromatic',
    path = 'chromatic.fs',
    send_vars = function (sprite)
        return {
            test = 123,
        }
    end
}
```

`chromatic.fs`
```
...
extern vec2 chromatic;
extern float test;
...
```

### DIFF
--- a/core/game_object.lua
+++ b/core/game_object.lua
@@ -2605,7 +2605,7 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
             'path',
         },
         set = 'Shader',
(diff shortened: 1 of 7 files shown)


stupxd:
don't merge this yet, there might be a miiiiiinor inconvenience

stupxd:
should be ok now

PR #263, opened 2024-09-19
dynamically display ante in collection during a round

stupxd:
also remove wonky spacing logic from vanilla


WilsontheWolf:
Just tested it and it seems to work good

PR #266, opened 2024-09-23
localization improvements

larswijn:
+ Translated multiple Chinese localization placeholders
+ Added placeholder entries for `b_stickers` and `b_graphics_mipmap_level` in both Spanish localizations - if someone can translate these, please do!
+ Cleaned up the indentation in all 4 localization files


larswijn:
Partial duplicate of #264

PR #274, opened 2024-10-08
Fix bp/bs messages

Eremel:
Blueprint and Brainstorm (and any custom jokers that use similar methods) display their status text on the card they are copying. These patches should fix this to make the message appear under the blueprint/brainstorm card instead.


stupxd:
Looks good

PR #279, opened 2024-10-17
safety check for rank

lshtech:
I've gotten the occasional crash on the enhancements collection for trying to read rank.face but rank is nil. I assume some weird mod interaction.


WilsontheWolf:
Not sure how this would happen, but doesn't seem like a bad idea.

PR #280, opened 2024-10-21
Fix restart when no mods are found

larswijn:
Steamodded would restart the entire game when closing the mods menu, because/if no mods were found.
Now checks if the delta mods counter actually exists before restarting.


WilsontheWolf:
Thats probably my mistake, whoops

PR #281, opened 2024-10-22
Enhancement clean ups

Eremel:
- Editions `extra_cost` value ignored `G.GAME.inflation`
- Enhancements that provide bonus chips have their values seperated from permanent bonus effects, such as Hiker
- Added `SMODS.poll_enhancement` *(follows same structure as other poll functions)*
- Renamed `name` in `SMODS.poll_seal` to `key` *(will break current uses of function with defined options as `{name:weight}` paired tables)*


Aurelius7309:
LGTM

PR #285, opened 2024-10-27
Deck Skins

onesuchkeeper:
Adds DeckSkin as a GameObject and provides an API for mods to add deck skins.

The API accepts:
- key - The identifying key.
- suit - The suit this DeckSkin applies to.
- ranks - The ranks this DeckSkin supports.
- lc_atlas - The name of the atlas used for the DeckSkin's low contrast texture.
- hc_atlas - Optional - The name of the atlas used for the DeckSkin's low contrast texture. If missing the lc_atlas will be used instead.
- posStyle - Optional - Determines the way cards in the textures are accessed. If 'collab' positions will be taken from G.COLLABS.pos, which matches the current layout of Balatro's collabs. If 'deck' positions will be taken from the card's pos, which matches the current layout of the deck texutre. if 'suit' the y position will be 0 and the x position will be taken from the card's pos. If missing will default to 'deck'.
- loc_txt - Optional - The translated name of the DeckSkin. If missing and a translation doesn't already exist, the key will be used.

--**Usage Example**--
SMODS.Atlas{
    key = "myLowContrast",
    path = "myLowContrast.png",
    px = 71,
    py = 95,
    atlas_table = 'ASSET_ATLAS'
}

SMODS.Atlas{
    key = "myHighContrast",
    path = "myHighContrast.png",
    px = 71,
    py = 95,
    atlas_table = 'ASSET_ATLAS'
}

SMODS.DeckSkin{
	key = "myDeckSkin",
	suit = 'Spades',
	ranks = {
		'2', '3', '4', '5', '6', '7', '8', '9', '10', 'Jack', 'Queen', "King", "Ace"
	},
	hc_atlas = "myHighContrast",
	lc_atlas = "myLowContrast",
	loc_txt = {
		['en-us'] = 'My Cool Deck Skin!'
	},
	posStyle = 'suit'
}

Also expands the 'Customize Deck' ui to show all suits, ranks and DeckSkins, including modded ones.

### DIFF
--- a/lovely/deck_skins.toml
+++ b/lovely/deck_skins.toml
@@ -5,7 +5,6 @@ priority = 0
 
 #========================================================#
 # Choose any rank for custom deck and use provided atlas #
-# and rework translation too I guess#
 #========================================================#
 [[patches]]
 [patches.regex]
@@ -83,6 +82,39 @@ until rank.next[1] == '2'
 
 for i = #cards, 1, -1 do
     face_cards:emplace(cards[i])
+end
+'''
+match_indent = true
+
+[[patches]]
+[patches.regex]
+target = "functions/UI_definitions.lua"
+pattern = '''function create_UIBox_customize_deck()([\s\S]*?)end'''
+position = "at"
+payload = '''
+function create_UIBox_customize_deck()
+  local suitTabs = {}
+
+  local index = 1
+  for suitKey, _ in pairs(SMODS.Suits) do
+    suitTabs[index] = {
+                label = localize(suitKey, 'suits_plural'),
+                tab_definition_function = G.UIDEF.custom_deck_tab,
+                tab_definition_function_args = suitKey
+            }
+    index = index + 1
+  end
+
+  suitTabs[1].chosen = true
+
+  local t = create_UIBox_generic_options({ back_func = 'options', snap_back = nil, contents = {
+    {n=G.UIT.R, config={align = "cm", padding = 0}, nodes={
+      create_tabs(
+        {tabs = suitTabs, snap_to_nav = true, no_shoulders = true}
+    )}}}
+  })
+
+  return t
 end
 '''
 match_indent = true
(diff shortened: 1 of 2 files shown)


lovely/deck_skins.toml
```diff
@@ -0,0 +1,119 @@
...
+position = "at"
+payload = '''
+function create_UIBox_customize_deck()
+  local suitTabs = {}
+
+  local index = 1
+  for suitKey, _ in pairs(SMODS.Suits) do
+    suitTabs[index] = {
+                label = localize(suitKey, 'suits_plural'),
+                tab_definition_function = G.UIDEF.custom_deck_tab,
```
Aurelius7309, line 101: There's a crash when selecting a custom suit with no deck skins registered for that suit (apparently due to some localization entries missing). Tested with Bunco

lovely/deck_skins.toml
```diff
@@ -0,0 +1,119 @@
...
+# Extend custom deck ui #
+#=======================#
+[[patches]]
+[patches.regex]
+target = "functions/UI_definitions.lua"
+pattern = '''local face_cards = CardArea\(([\s\S]*?)\)'''
+position = "at"
+payload = '''
+
+local rankCount = 0
```
Aurelius7309, line 52: Steamodded wants to preserve the behavior of the game for aspects that other mods don't actively change. As such, I can't approve always putting all ranks on display. At the very least the vanilla collabs should only display the face cards they're changing. Maybe you could make use of the `ranks` table to respect that?

Aurelius7309 (changes requested):
Just tested and I have a few remarks. Looks good to merge otherwise once these are taken care of

lovely/deck_skins.toml
onesuchkeeper: Only ranks with skins will be shown now. This occurs though all skins, so a '3' of Spades texture on 'MySkin' will cause a 3 to show up in all other Spade skins even without a 3. The vanilla deck skin tabs do not update the ranks they display between skins, only suits which causes this. I could force an update, but with the way the 'default' skin matches the collab skin ranks I believe this is the intent and fully shows the differences beetween skins.

lovely/deck_skins.toml
onesuchkeeper: Crash fixed

Aurelius7309 (approved):
LGTM

PR #295, opened 2024-11-16
Add coroutine to intertwine updates/gc cycles with scoring

cg-223:
This eliminates the memory overhead of scoring, and eliminates the issue of the game freezing while scoring is in progress.

An overlay with a simple "Calculating..." text appears while scoring, and the user has no control while this is happening.

The overlay only appears during the actual calculations involved in scoring.

The scoring animations the user sees are outside of the calculations, and as such the obstruction of the overlay disappears afterwards.

Updates run 10 times a second (along with GC cycles). If scoring takes less than a tenth of a second, the effects of the coroutine are nonexistent.

It's to be noted that scoring will always take up at least 0.1 seconds. The GC cycle time isn't included.

It seems that updates take a shit ton of time if scoring animations are enabled, likely due to the event manager iterating through all of the events in queue. This is negligible, since if you're performing that many retriggers with scoring animations on, the time it'll take to fully run through all the animations would be months anyways.

### DIFF
--- a/core/overrides.lua
+++ b/core/overrides.lua
@@ -2067,9 +2067,63 @@ function get_pack(_key, _type)
    if not center then center = G.P_CENTERS['p_buffoon_normal_1'] end  return center
 end
 
+--scoring coroutine
+local oldplay = G.FUNCS.evaluate_play
+
+function G.FUNCS.evaluate_play()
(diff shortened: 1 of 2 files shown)


cg-223:
moving this PR to talisman

PR #305, opened 2024-11-28
"Quantum" Enhancements

MathIsFun0:
Allows Jokers (and other effects, theoretically) to make playing cards act as if they have multiple enhancements.

Features:
- SMODS.has_enhancement(card, key)
- SMODS.get_enhancements(card, extra_only)
- Works with enhancement gate checks
- Works with vanilla jokers and most vanilla enhancements

Todo:
- Compatibility with wild cards and similar effects
- Full compatibility with enhancement API


nebulazorua:
In this same line, could be useful to have something like is_id() for quantum IDs, Incase you wanted to make, say, a Joker or Enhancement that makes a card count as multiple ranks.

Aurelius7309:
There's been discussions on the topic of quantum ranks, it's turned out quite complicated to handle straights efficiently if that is implemented, though the rank checking process would have to be normalized for other ranks as well while respecting custom ones. That's a pretty big blocker I think, but feel free to have a go at it

nebulazorua:
> There's been discussions on the topic of quantum ranks, it's turned out quite complicated to handle straights efficiently if that is implemented, though the rank checking process would have to be normalized for other ranks as well while respecting custom ones. That's a pretty big blocker I think, but feel free to have a go at it

I see. Might give it a look some day once I have some more free time

MathIsFun0:
updated my todo list
there is also a bug with cards being saved with their "quantum" enhancements

MathIsFun0:
this is feature complete I think now
unless we want to change the way things like quantum stone and wild cards work

Aurelius7309 (changes requested):
Looks fine for the most part, but I've noticed stone/no_suit/no_rank enhancements are unsupported as extra enhancement due to checking for just the main center. Also how should cards that have both all_suits and no_suit behave? this is relevant for poker hand evaluation

MathIsFun0:
wild stone cards are cursed now
image

Aurelius7309 (approved):

PR #306, opened 2024-11-28
Ignore disabled mods when evaluating dependencies

DigitalDetective47:
Previous behaviors caused mods to be unable to load if the conflicting mod was in the mods folder, even if it was disabled. This made having multiple mod groups installed at the same time inconvenient.


DigitalDetective47:
No, `check_dependencies` displays a conflict even if the conflicted mod is disabled in the mods menu.

wingedcatgirl:
image

The disabled mod is indeed considered an unresolved conflict.

Aurelius7309:
> image
> 
> The disabled mod is indeed considered an unresolved conflict.

This is true only of legacy headers (which are deprecated and not getting any changes made to them) but fixed with JSON metadata

PR #309, opened 2024-11-29
Individual Undiscovered Sprite

GauntletGames-2086:
Allows centers affected by `SMODS.UndiscoveredSprite` to ignore it and use their own custom undiscovered sprite handling, using the `undiscovered` table.


Aurelius7309:
lgtm for the most part, there's a typo (no_overly) in line 91 though. Suggestion: should we provide a fallback undiscovered sprite to avoid crashes when it's missing?

PR #310, opened 2024-11-29
Update real_card_limit when creating negative (playing) card

badfontkeming:
Fixes situations where creating negative cards results in a negative handsize after playing or discarding them

Repro case:

- Start a new game
- Use debug tools to mark a card as negative. Hand size is now 9/9 as a new card is drawn.
- Discard the negative card. Hand size is now 8/8 as expected.
- Make another negative card and discard it. Hand size is now 8/7. Hand size is permanently reduced.

This appears to be happening because discarding a negative card updates real_hand_size, where creating a new negative card doesn't.


Eremel:
Looks good, merged.

PR #314, opened 2024-12-03
Fix DeckSkin outdated false flag

Kekulism:
Fixed mods that use DeckSkin being falsely flagged as outdated

### DIFF
--- a/loader/loader.lua
+++ b/loader/loader.lua
@@ -55,7 +55,7 @@ function loadMods(modsDirectory)
         },
         prefix        = { pattern = '%-%-%- PREFIX: (.-)\n' },
         version       = { pattern = '%-%-%- VERSION: (.-)\n', handle = function(x) return x and V(x):is_valid() and x or '0.0.0' end },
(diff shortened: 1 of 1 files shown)


loader/loader.lua
```diff
@@ -55,7 +55,7 @@ function loadMods(modsDirectory)
         },
         prefix        = { pattern = '%-%-%- PREFIX: (.-)\n' },
         version       = { pattern = '%-%-%- VERSION: (.-)\n', handle = function(x) return x and V(x):is_valid() and x or '0.0.0' end },
-        outdated      = { pattern = { 'SMODS%.INIT', 'SMODS%.Deck' } },
+        outdated      = { pattern = { 'SMODS%.INIT', 'SMODS.Deck[:.]new' } },
```
WilsontheWolf, line 58: the %. is nessicary as . represents any character, and the % escapes it

WilsontheWolf (changes requested):

loader/loader.lua
Kekulism: Like this?

WilsontheWolf (approved):

PR #330, opened 2024-12-14
Allow String keys in Localization Strings and Add default loc_vars function

Cheesy-Brik:
Adds a default loc_vars function and allows for returned args from loc_vars to use not just number keys but also string keys. Alters the localization function directly so the string key change can be used in any text field as long as the args has that string key.

Example:
```lua
SMODS.Joker {
    key = 'pacifist',
    loc_txt = {
        name = 'Pacifist Joker',
        text = {"Played cards with", "{C:spades}#suit#{} suit give", "{C:chips}+#chips#{} Chips when scored"}
    },
    config = {
        chips = 50,
        suit = 'Spades'
    },
...
```

Would remove a lot of boilerplate and would probably mean most people would never have to interact with the loc_vars function

It also reads the `extra` table as a fallback if the key doesn't exist.

### DIFF
--- a/lovely/localization.toml
+++ b/lovely/localization.toml
@@ -8,7 +8,7 @@ priority = 0
 target = 'functions/misc_functions.lua'
 pattern = '''args\.vars\[(?<index>.*)\]'''
 position = 'at'
-payload = '''args.vars[$index] or args.vars.extra[$index]'''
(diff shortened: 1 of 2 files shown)


Aurelius7309:
Not needed or wanted in terms of directing new modders towards learning concepts properly, as discussed on Discord.

PR #334, opened 2024-12-18
Correct mentions of LÖVE

Keyslam:
Sorry if this comes off as nitpicky or annoying, but I feel like it's important to get right:

The framework is called LÖVE, and the url for it is love2d.org. 'LÖVE2D' and 'Love2D' aren't right, nor is it an engine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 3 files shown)


Aurelius7309:
Thanks for clarifying :v

PR #348, opened 2024-12-23
Fix hand size reduction caused by negative playing cards in tarot packs

badfontkeming:
Whoops. Should've caught this the first time I fixed issues with this part of the code. Would've sent this in sooner but life got in the way.

Repro:

- Open a tarot or spectral pack
- Make a playing card negative (DebugPlus or Cryptid's 'Ritual' spectral card)
- Hand size is permanently reduced by 1


Aurelius7309:
whoops... though as a remark the code here should also check for custom booster packs that draw cards

badfontkeming:
Yep, probably so. I can confirm that it'll draw a card in one of Pokermon's booster packs.

I won't lie, I'm not really familiar with Steamodded's API at all, so I don't know what check would be appropriate to accomplish that.

Aurelius7309:
That's fine, I can add it myself

badfontkeming:
Much appreciated.

PR #355, opened 2024-12-26
Better Calc 2.0

Eremel:
### NOTE: this WILL break current enhancements, and possibly seals too

Moves enhancement calculation to use the same system as all other calculations.

Enhancements and seals now use the same effect evaluation in `G.FUNCS.evaluate_play`, can easily accept new modifications if they are added to `eval_card` correctly.
Use `if context.cardarea == G.hand or G.play and not context.repetition` to calculate enhancements/seals. Valid returns in are `chips`, `mult`, `p_dollars`, `dollars`, `x_mult`, `func`. Messages are automatically created for the scoring card, but specifying `card = X` will change where it appears, or use `message` to add a custom message in addition. To disable the automatic message, you can add `remove_default_message = true` to the return table of the object.
Use `if context.cardarea == G.hand or G.play and context.repetition_only` to calculate repetitions. Return is
```lua
{
   repetitions = X,
   card = card,
   message = localize('k_again_ex')
}
```

Also adds the `destroying_card` context to scored enhanced/sealed cards.


Eremel:
**This will break a lot of enhancement and edition calculations, maybe some other stuff too 🙃**

Updated entire scoring calculation. All card evaluation now uses `eval_card` to allow all object types to be considered in scoring.
All calculation contexts are now passed to a card's edition, enhancements, seal and stickers.

Refactored Retrigger API *(fingers crossed there are no bugs here)* which is now more efficient. Have not incorporated levels of retriggering as of yet.

Expanded all contexts to use `SMODS.calculate_context`, a function that runs the context through `G.jokers`, `G.consumeables`, `scoring_hand` *(if present)*, `G.hand` and `G.GAME.selected_back` and calculates effects. This function can easily be hooked to add different areas to calculate over. **Main scoring evaluation does not use this function**

Adds `SMODS.calculate_individual_effect` to handle evaluate effects returned from `calculate` functions. Aims to unify key returns *(returning `chips = X` from ANY context is accepted)* and adds support for overwriting the default message. Again, this function is easily hooked to add different scoring keys, such as `x_chips` from Talisman, etc.

`SMODS.eval_this` now increases `percent` to change the pitch of sound triggers throughout scoring.

Added support to return a `sound = key` in calculation to overwrite whatever sound `card_eval_status_text` normally uses.

*I think buried in here somewhere is support for `keep_on_use` consumables in booster packs*

src/utils.lua
```diff
@@ -908,6 +950,240 @@ SMODS.find_mod = function(id)
...
+-- Used to calculate contexts across G.jokers, scoring_hand (if present), G.play and G.GAME.selected_back
+-- Hook this function to add different areas to MOST calculations
+function SMODS.calculate_context(context, percent, return_table)
+    context.cardarea = G.jokers
+    for k=1, #G.jokers.cards + #G.consumeables.cards do
+        local _card = G.jokers.cards[k] or G.consumeables.cards[k - #G.jokers.cards]
+        --calculate the joker effects
+        local effects = {eval_card(_card, context)}
+        if effects[1].retriggers then
+            context.retrigger_joker_check = true
```
MathIsFun0, line 1144: This is where the actual retriggering is, so context.retrigger_joker should be what is set to true here.

MathIsFun0 (changes requested):
There appears to be no indicator that a joker is successfully triggered.
Retrigger API only retriggers a joker if it successfully triggers the first time, hence why the version in better calc is much slower.
This is also how context.post_trigger works - it's calculated after a joker is successfully triggered.

The way I set up checking if a joker is succesfully triggered with the old calculation system is by making it return a value, or "nil, true" if it calculated without returning a value.

Note that joker_retrigger.toml held a bunch of patches that gave vanilla jokers compatibility with trigger detection; these should be either merged into better_calc.toml or put in a new file. joker_retrigger.toml should be deleted now that it's not being used (the one patch in there right now is useless).

Besides that, the code looks fine from my quick glance, although I didn't try testing it with some mods because of the joker retrigger issues.

Aurelius7309 (approved):

PR #373, opened 2025-01-06
Fix Better Calc 2.0

GauntletGames-2086:
Current discovered issues (check = resolved): 
- [x] Patches are not landing. 
- [x] Crash at end of round when joker gives a payout. 
- [x] Foil and Holo do not give chips/mult. 
- [x] Gold Cards do not give dollars. 
- [x] Mult goes before chips on enhancement (e.x. Mult Card). 
- [x] Jokers using context.individual do not juice cards correctly. In vanilla, mult message appears on top of card and both juice. SMODS now displays mult below joker, and playing card does not juice. 
- [x] Stone Card and Wild Card do not work
- [x] Red Seal's "Again!" shows up regardless of if it was played
- [x] Purple Seal will give two tarot cards instead of one
- [x] Blueprint cancels any non-blueprintable calculation that happens after it


GauntletGames-2086:
sorry I forgor that there's simply WAAAY too much broken

GauntletGames-2086:
All currently known issues have been resolved, and any future ones are (likely) to not be destructive enough to just fix on main. AKA ready to merge.

PR #383, opened 2025-01-09
Fix copy-paste issue with lovely patch targetting matador

larswijn:
Relevant discord message: https://discord.com/channels/1116389027176787968/1209564621644505158/1327008498982064240


larswijn:
Already implemented

PR #389, opened 2025-01-10
Change straight algorithm

Aurelius7309:
The current algorithm for straights is pretty bad with custom ranks that have split paths. This proposes a new one that works correctly in cases where the previous one failed.
Example: Given ranks 10->11->12->13 with `SMODS.Ranks['10'].next = {'Jack', '11'}`.
- 9,10,J,Q is a valid 4-finger straight
- 9,10,J,11,12 should also be a valid 4-finger straight but isn't because the current algorithm gets trapped in the first path

Originally tried to base an implementation of quantum ranks on this but realized this would be inefficient as candidate straights can be invalidated at each step and checking a given hand with quantum ranks for containing a straight is exponential in the length of the straight.

Also builds prev tables in ranks when injected. Not used here but figured it might be useful to have

### DIFF
--- a/src/game_object.lua
+++ b/src/game_object.lua
@@ -1781,6 +1781,7 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
             ignore = false
         },
         next = {},
+        prev = {},
         straight_edge = false,
         -- TODO we need a better system for what this is doing.
         -- We should allow setting a playing card's atlas and position to any values,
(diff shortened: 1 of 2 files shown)


Aurelius7309:
Performance is about the same, shortcut is slower with large amounts of cards selected. Measured around 2.5ms with shortcut vs. 0.8ms without, while the previous implementation takes barely any hit at all

DigitalDetective47:
are `prev` tables built automatically by reversing the `next` tables, or do they have to be specified manually?

Aurelius7309:
Builds prev tables automatically, but I didn't end up using them here

PR #393, opened 2025-01-11
More deck contexts, SMODS.Back.calculate

Aurelius7309:
Adds context.individual and context.other_joker/other_consumeable as evaluated contexts for decks. Also introduces `SMODS.Back.calculate` and deprecates `SMODS.Back.trigger_effect` to indicate consistency with other calculate functions. This calculate function as well as `SMODS.Back.apply` get passed the `Back` they're called on, again for consistency


Aurelius7309:
Apparently context.main_scoring has inconsistent context args between deck and play areas?

PR #396, opened 2025-01-11
fix: play card referenced when destroying hand cards

baliame:
This patch adds calculations for destroying cards in hand when a hand is played, but the if block refers to play cards.


Eremel:
Fixed in 1311g

PR #411, opened 2025-01-16
BetterCalc 2.1:tm:

Aurelius7309:
- Adds util functions for a lot of code that was previously patched into state_events.lua. This should further simplify adding calculation to custom card areas. Added functions are `score_card`, `calculate_main_scoring`, `calculate_end_of_round_effects`, `calculate_destroying_cards`.
- Adds SMODS.optional_features, which can be enabled by defining a `optional_features` function (or table, in which case calling it is skipped). Currently available options are `cardareas.deck` and `cardareas.discard` to enable scoring calculation for the respective areas.
- Adds `context.main_eval` to calculation that affects jokers directly and doesn't have a secondary card, replaces `not context.individual and not context.repetition`.


Aurelius7309:
Note: this includes the enhancement cache commit because i started working on this using the wrong branch. Its code has been removed because it has not been fully reviewed, which I'm planning to do later.

PR #412, opened 2025-01-16
Implement enhancement cache

Aurelius7309:
Remaking #408 because i accidentally worked on top of it


Lordmau5:
Ran into some performance issues with old-calc, 301 Lucky Cards and Driver's License as well (together with Cryptid and a few other mods)

Implemented similar caching as well, managed to get a lot of FPS back.

I would highly suggest also implementing it into the `SMODS.has_enhancement` function, too, seeing as that is used quite a few times as well (and also has a call to `calculate_joker` on every Joker)

PR #416, opened 2025-01-18
Fix: Gold cards are triggering twice

baliame:
version: **1.0.0~ALPHA-1318d-STEAMODDED**

Investigated this issue, I've managed to narrow it down to eval_card being called twice with identical contexts: `context.end_of_round = true, context.playing_card_end_of_round = true`

I think I have the root cause.

First evaluation (normal), will calculate card's "regular" enhancement: smods/utils.lua:1394

`local effects = {eval_card(card, context)}`

Second evaluation (extra), will calculate card's extra enhancement: smods/utils.lua:1403

`effects[#effects+1] = eval_card(card, context)`

This should not be called at all, the card is a regular gold card with the quantum enhancement feature on default setting (turned off). However, it is called, because smods/utils.lua:1395

`local extra_enhancements = SMODS.get_enhancements(card, true)`

Returns an enhancement (the card's base enhancement) despite being called with extra_only.

I'm not sure what the intention is here so I'm not confident the patch is correct in every case.

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -851,7 +851,10 @@ function Card:calculate_enhancement(context)
 end
 
 function SMODS.get_enhancements(card, extra_only)
-    if not SMODS.optional_features.quantum_enhancements then return card.ability.set == 'Enhanced' and { [card.config.center.key] = true } or {} end
+    if not SMODS.optional_features.quantum_enhancements then
+	if extra_only then return {} end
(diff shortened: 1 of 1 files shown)


baliame:
Seems like I didn't have the version I thought I had checked out and it was fixed in the meantime, my bad.

Aurelius7309:
The original commit was a hasty fix to a previously inverted check. I'm gonna change this one inline but thanks for pointing it out

PR #443, opened 2025-02-05
fix repetitions ignoring `remove_default_message`

DigitalDetective47:
[1404e](https://github.com/Steamodded/smods/commit/1189481445a3ddef598c978d67df78d0ac8e4ed6) added a default message for repetetions and joker retriggers. unfortanately, this default message is always applied, even if `remove_default_message` is specified. this pull request fixes that bug.

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -1216,7 +1216,7 @@ SMODS.calculate_repetitions = function(card, context, reps)
         if value.repetitions then
             for h=1, value.repetitions do
                 value.card = value.card or card
-                value.message = value.message or localize('k_again_ex')
+                value.message = value.message or (not value.remove_default_message and localize('k_again_ex'))
                 reps[#reps+1] = {key = value}
             end
         end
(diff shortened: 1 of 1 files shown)


DigitalDetective47:
an example of this can be seen in the joker [Forty-seven](https://github.com/DigitalDetective47/strange-pencil/blob/60369521ef566381c16d90faacfea615169c3984/jokers.lua#L226) from [Strange Pencil](https://github.com/DigitalDetective47/strange-pencil). SMODS 1404e caused it to display a duplicate Again! message, as it had assumed there was not a default message. (it displays the Again! message directly in the `func` in order to make it happen at the desired time) i attempted to fix this issue in [0.3.0~0b05d](https://github.com/DigitalDetective47/strange-pencil/commit/25321ece03440314e164449e89c16ec074ad869c) by adding the `remove_default_message` parameter, but it doesn't work because of this bug. when using the pr version, you can see that only one Again! message appears.

PR #453, opened 2025-02-10
Fix set_edition not properly resetting negative playing cards

SylviBlossom:
This fixes hand size permanently increasing by 1 when using Death to turn a negative card non-negative.
(just a copy of the addition code later down in the function, but subtracting instead)


Eremel:
I cannot reproduce this issue locally, what mods are you running?

SylviBlossom:
i have tested with just DebugPlus enabled to turn playing cards negative and nothing else, it works as expected with no issues (maintains standard hand size of 8) until using Death to remove negative, where the `real_card_limit` stays the same despite that

the issue can probably only show up once a negative card has been naturally drawn/removed from the hand since that's when the `real_card_limit` is initialized

PR #459, opened 2025-02-14
Add french localization

Loloweb:
Thought I might as well add localization for this since I'm working on another mod.
Based on Balatro's official localization, meaning words like "seed" aren't translated. "Mipmap" also doesn't have a real french equivalent.


Loloweb:
whoops, it should be correctly formatted now

Aurelius7309:
Thanks!

PR #462, opened 2025-02-15
fix seed advancing on collection

Toneblock:
noticed a slight flaw where calling a pseudorandom in set_ability advances seeds while in the collection

kept this behaviour for to-do list, which, despite an attempted "fix" visible in card.lua, still has this issue in vanilla


Toneblock:
not sure if this is the correct way to fix this but it seems to work

PR #469, opened 2025-02-19
Shop utils

Eremel:
Adds support for modifying voucher count, booster count, free reroll count in shops. Adds a max width for the shop to stop increasing the size extending off the screen.

`SMODS.change_voucher_limit(mod)` changes the voucher limit by specified value. If in shop adds the correct number of vouchers to the current shop
`SMODS.add_voucher_to_shop(key)` adds a voucher to the current shop (if no key provided a random voucher is selected)
`SMODS.change_booster_limit(mod)` changes the booster limit by specified value. If in shop adds the correct number of boosters to the current shop
`SMODS.add_booster_to_shop(key)` adds a booster to the current shop (if no key provided a random booster is selected)
`SMODS.change_free_rerolls(mod)` changes the number of free rerolls. If in shop added immediately


srockw:
It'd be nice if the `add` functions returned the card they created, so you're able to modify them afterwards, for example, to change the voucher/booster price.

Aurelius7309 (approved):
Looks good to me

PR #470, opened 2025-02-20
feat: cross-platform https

WilsontheWolf:
This adds a WIP implementation of the [https](https://love2d.org/wiki/lua-https) module that works cross platform without having to ship native libs. It uses one of 2 backends (the actual https module for windows, and libcurl for other platforms).

Basic example
```lua
local https = require "SMODS.https"

print(https.request("https://example.com"))
```

API differences from upstream lua https:
- Ships with a default user agent
- Additional checks on some values
- Proper error messages (curl backend only)

This has been tested on the following systems:
- Windows (via wine)
- Linux native build (Fedora 41)
- MacOS (Ventura intel)

Needs additional testing
- MacOS (No Xcode, No brew or other package manager) (Need to check if curl is actually provided by macos)
- ~~iOS (Not an offically support platform)~~ Doesn't work

If MacOS doesn't work, I can implement a version using objc api's. 

I also want to implement an a function to make a request on a background thread, but that will come later (possibly other pr).

TODO Before Merge:
- [x] Make changes for luals
- [x] Make wiki page
- [x] ~~Get someone with a MacOS device not setup for dev stuff to test~~ (fuck it we ball)
- [x] (optional) Async implementation

### DIFF
--- a/libs/https/luajit-curl.lua
+++ b/libs/https/luajit-curl.lua
@@ -1,0 +1,1056 @@
+-- This is a modifed version of https://gist.github.com/LPGhatguy/09d3955207ab35d96e97
+-- Original license:
+--[[
+LuaJIT-cURL
+Lucien Greathouse
+LuaJIT FFI cURL binding aimed at cURL version 7.38.0.
+
+Copyright (c) 2014 lucien Greathouse
+
+This software is provided 'as-is', without any express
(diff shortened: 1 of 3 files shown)


Aurelius7309 (approved):
lgtm

PR #479, opened 2025-02-24
Rewrite set_blind to respect overriding vanilla blinds

baliame:
There's a stupid amount of patches because it's not even an elseif chain for some reason.


baliame:
A few refactors of this refactor later it should be ready.

PR #480, opened 2025-02-24
`SMODS.DrawStep` Card.draw modularization

Aurelius7309:
Card.draw is way too annoying to deal with, so why not build an API around making it modular and allowing easy insertions anywhere during the drawing process just by specifying a numerical priority value? Well here it is I guess


src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+        self.ARGS.send_to_shader[2] = G.TIMERS.REAL
+
+        for k, v in pairs(self.children) do
+            v.VT.scale = self.VT.scale
+        end
+    
+        G.shared_shadow = self.sprite_facing == 'front' and self.children.center or self.children.back
+    
+        --Draw the shadow
+        if not self.no_shadow and G.SETTINGS.GRAPHICS.shadows == 'On' and((self.ability.effect ~= 'Glass Card' and not self.greyed and self:should_draw_shadow() ) and ((self.area and self.area ~= G.discard and self.area.config.type ~= 'deck') or not self.area or self.states.drag.is)) then
```
WilsontheWolf, line 37: Could we do the no_shadow and shadows on check earlier in this function to avoid doing some extra work we don't need to?

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+            if (self.config.center.undiscovered and not self.config.center.undiscovered.no_overlay) or not( SMODS.UndiscoveredSprites[self.ability.set] and SMODS.UndiscoveredSprites[self.ability.set].no_overlay) then 
+                shared_sprite:draw_shader('dissolve', nil, nil, nil, self.children.center, scale_mod, rotate_mod)
+            else
+                if SMODS.UndiscoveredSprites[self.ability.set] and SMODS.UndiscoveredSprites[self.ability.set].overlay_sprite then
+                    SMODS.UndiscoveredSprites[self.ability.set].overlay_sprite:draw_shader('dissolve', nil, nil, nil, self.children.center, scale_mod, rotate_mod)
+                end
+            end
+        end
+
+        if self.ability.name == 'Invisible Joker' and (self.config.center.discovered or self.bypass_discovery_center) then
```
WilsontheWolf, line 156: Should we make a property that allows other cards to be able to get this behviour?

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+}
+
+SMODS.DrawStep {
+    key = 'front',
+    order = 0,
+    func = function(self, layer)
+        if not self.vortex and self.sprite_facing == 'front' then
+            --Draw the main part of the card
+            if (self.edition and self.edition.negative) or (self.ability.name == 'Antimatter' and (self.config.center.discovered or self.bypass_discovery_center)) then
+                if self.children.front and (self.ability.delayed or (self.ability.effect ~= 'Stone Card' and not self.config.center.replace_base_card)) then
```
WilsontheWolf, line 176: Same as above for invisible joker

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+        end
+    end,
+}
+
+SMODS.DrawStep {
+    key = 'seal',
+    order = 30,
+    func = function(self, layer)
+        if not self.vortex and self.sprite_facing == 'front' then
+            local seal = G.P_SEALS[self.seal or {}] or {}
```
WilsontheWolf, line 253:
```suggestion
            local seal = G.P_SEALS[self.seal] or {}
```

I don't believe that is nessicary

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+SMODS.DrawStep {
+    key = 'floating_sprite',
+    order = 60,
+    func = function(self)
+        if not self.vortex and self.sprite_facing == 'front' then
+            if self.config.center.soul_pos and (self.config.center.discovered or self.bypass_discovery_center) then
+                local scale_mod = 0.07 + 0.02*math.sin(1.8*G.TIMERS.REAL) + 0.00*math.sin((G.TIMERS.REAL - math.floor(G.TIMERS.REAL))*math.pi*14)*(1 - (G.TIMERS.REAL - math.floor(G.TIMERS.REAL)))^3
+                local rotate_mod = 0.05*math.sin(1.219*G.TIMERS.REAL) + 0.00*math.sin((G.TIMERS.REAL)*math.pi*5)*(1 - (G.TIMERS.REAL - math.floor(G.TIMERS.REAL)))^2
+
+                if self.ability.name == 'Hologram' then
```
WilsontheWolf, line 321: Same as the others above although this one seems less likely to want to be used

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+            end
+        end
+    end,
+}
+
+SMODS.DrawStep {
+    key = 'back',
+    order = 0,
+    func = function(self)
+        if not self.vortex and self.sprite_facing == 'back' then
```
WilsontheWolf, line 377: Putting this here as it's where I thought of it but it might be nice to add a propert to the draw step to skip this step in certain conditions. Something like 
```lua
skip = {
	vortex = true,
	front = true,
}
```
which would just be equivalent to putting the code in a big old 
```lua
if not self.vortex and self.sprite_facing ~= "front" then
end
```

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+SMODS.DrawStep {
+    key = 'back_sticker',
+    order = 10,
+    func = function(self)
+        if not self.vortex and self.sprite_facing == 'back' then
+            if self.sticker and G.shared_stickers[self.sticker] then
+                G.shared_stickers[self.sticker].role.draw_major = self
+                local sticker_offset = self.sticker_offset or {}
+                G.shared_stickers[self.sticker]:draw_shader('dissolve', nil, nil, true, self.children.center, nil, self.sticker_rotation, sticker_offset.x, sticker_offset.y)
+                if self.sticker == 'Gold' then G.shared_stickers[self.sticker]:draw_shader('voucher', nil, self.ARGS.send_to_shader, true, self.children.center) end
```
WilsontheWolf, line 406: Same as the other's

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+        end
+    end,
+}
+
+SMODS.DrawStep {
+    key = 'others',
+    order = 90,
+    func = function(self)
+        for k, v in pairs(self.children) do
+            if not v.custom_draw and k ~= 'focused_ui' and k ~= "front" and k ~= "back" and k ~= "soul_parts" and k ~= "center" and k ~= 'floating_sprite' and k~= "shadow" and k~= "use_button" and k ~= 'buy_button' and k ~= 'buy_and_use_button' and k~= "debuff" and k ~= 'price' and k~= 'particles' and k ~= 'h_popup' then v:draw() end
```
WilsontheWolf, line 417: This code would probably be cleaner with a table to lookup in and then mods could also add their own values to the table to not be drawn here if they want their own step.

src/card_draw.lua
```diff
@@ -0,0 +1,446 @@
...
+        end
+    end,
+}
+
+SMODS.DrawStep {
+    key = 'focused_ui_2',
+    order = 100,
```
WilsontheWolf, line 424: Why do these need 2 different priorities?

WilsontheWolf (changes requested):

src/card_draw.lua
Aurelius7309: Indeed it is not

src/card_draw.lua
Aurelius7309: Not entirely sure why it wants the focused UI on top for cards in G.hand as I don't think it makes a difference for vanilla, but this is just copied from what vanilla is doing

src/card_draw.lua
Aurelius7309: All that would possibly be affected there is setting G.shared_shadow as the rest should happen regardless of that condition. Also some of it is just setup for shader variables but apparently it's not needed for some reason when running with `layer == 'card'`, not sure what that means either

src/card_draw.lua
Aurelius7309: I don't think that's needed here, as this step provides centers with a draw function to do this and I don't think the voucher shader deserves any special treatment just because it's used by invis

src/card_draw.lua
Aurelius7309: That one makes a bit more sense maybe as there isn't a direct way for anything but suits and ranks to intercept here

src/card_draw.lua
Aurelius7309: Using hologram's code in particular doesn't seem too useful, but maybe there should be a soul_pos.draw that allows customization of how the floating sprite is drawn

src/card_draw.lua
Aurelius7309: I might prefer stating this the other way around as a list of conditions, but something along those lines seems useful for sure

src/card_draw.lua
Aurelius7309: Actually this is already included as per replace_base_card

src/card_draw.lua
```diff
@@ -0,0 +1,453 @@
...
+    func = function(self)
+        if self.config.center.soul_pos and (self.config.center.discovered or self.bypass_discovery_center) then
+            local scale_mod = 0.07 + 0.02*math.sin(1.8*G.TIMERS.REAL) + 0.00*math.sin((G.TIMERS.REAL - math.floor(G.TIMERS.REAL))*math.pi*14)*(1 - (G.TIMERS.REAL - math.floor(G.TIMERS.REAL)))^3
+            local rotate_mod = 0.05*math.sin(1.219*G.TIMERS.REAL) + 0.00*math.sin((G.TIMERS.REAL)*math.pi*5)*(1 - (G.TIMERS.REAL - math.floor(G.TIMERS.REAL)))^2
+
+            if self.ability.name == 'Hologram' then
+                self.hover_tilt = self.hover_tilt*1.5
+                self.children.floating_sprite:draw_shader('hologram', nil, self.ARGS.send_to_shader, nil, self.children.center, 2*scale_mod, 2*rotate_mod)
+                self.hover_tilt = self.hover_tilt/1.5
+            elseif type(self.config.center.soul_pos.draw) == 'function' then
```
GauntletGames-2086, line 329: This and the Hologram checks should be swapped, so in the edgecase of someone taking ownership of Hologram to change the soul sprite `soul_pos.draw` is respected.

WilsontheWolf (approved):

GauntletGames-2086 (approved):
Besides one issue, this PR looks good to merge.

PR #486, opened 2025-02-25
feat: Blind descriptions in info_queue

baliame:


Aurelius7309 (dismissed):
This does not adhere to the specification of `Blind:collection_loc_vars`. 
grafik

baliame:
I see what you mean now. This should be spec-compliant.

PR #487, opened 2025-02-25
feat: Add permanent playing card bonuses

baliame:
Code mostly based on AMM (https://github.com/AutumnMood924/AutumnMoodMechanics)

Adds most values as permanent card bonuses similar to Hiker's chip bonus. The justification here is that both my mod and Ice's Kino (https://github.com/icyethics/Kino) mod implement parts of this system. Ice planned to transition to AMM as a dependency for this, but unless I also pull in AMM as a dependency just for this feature, our Card function overrides would run twice, essentially doubling every effect. Since the demand for the feature seems high, this seems like a good candidate to support on an SMODS level.

### DIFF
--- a/localization/en-us.lua
+++ b/localization/en-us.lua
@@ -62,7 +62,58 @@ return {
                     'This mod has been',
                     '{C:attention}disabled!{}'
                 }
-            }
+            },
+
+
+            -- card perma bonuses
+            card_extra_xchips = {
+                text = {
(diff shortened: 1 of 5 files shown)


VictinZero:
@larswijn had already started working on this and was relying on me to finish the implementation for Jokers. At the very least, we had discussed some conventions for both naming and timing which might be good to go over

PR #490, opened 2025-02-27
feat: Add card permanent bonuses (multiplicative stacking)

baliame:
#487 but more intuitively perma X effects stack multiplicatively


baliame:
Per discussion with Larswijn, we consider this feature ready now.

PR #493, opened 2025-02-27
fix: voucher crashes when indexing string as table

SynicalMX:
This PR addresses 2 issues that stem from the Multiplayer mod. I currently don't know where the root of the issue lies. But from the stack trace it was clear it was a smods issue, but please correct me if I'm wrong.

When spawning vouchers after a new round, it tries to index 'G.GAME.current_round.voucher' assuming it's a table, but from my testing it can either be a string, table, or just nil. This is now fixed to account for the different types.

I also noticed a crash for when indexing 'G.GAME.current_round.voucher.spawn', so it now replaces the pairs() call with an empty table if nil.

I am here for any questions/concerns.

### DIFF
--- a/lovely/shop.toml
+++ b/lovely/shop.toml
@@ -64,15 +64,22 @@ end
 '''
 payload = '''
 local vouchers_to_spawn = 0
-for _,_ in pairs(G.GAME.current_round.voucher.spawn) do vouchers_to_spawn = vouchers_to_spawn + 1 end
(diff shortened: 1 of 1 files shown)


Aurelius7309:
I believe the underlying issue in multiplayer was fixed already? Either way if a mod causes this value not to adhere to the new format Steamodded is setting, I would much rather have the game end up crashing than to silently ignore string values simply because it's better to debug. Of course it could be changed to work with the old format as well, but that seems unnecessarily likely to cause incompatibilities.

PR #500, opened 2025-03-01
fix unscored cards being unhighlighted

Toneblock:
when unscored card areas are enabled, unscored cards are unhighlighted despite never being highlighted
my ears don't like it
again, i'm not sure if this is how you fix it


MathIsFun0:
Shouldn’t this be done on Steamodded’s end? (If it happens without Cryptid)

Eremel:
Fixed in #d641505

PR #503, opened 2025-03-01
Fix perma_x_chips arithmetic crash

baliame:
I guess I never tested this on a resumed run.

## Additional Info:


Aurelius7309:
Doesn't this only cause issues when continuing a run started on an older version? That's not something we need to support and such runs will probably be broken in other ways. I'm thinking about adding a safety mechanism that warns when trying to continue a run started on an older version though

baliame:
Yeah, this is purely for resuming runs - up to you.

Aurelius7309:
Alright we don't need this then. I'll come up with something for the frontend

PR #508, opened 2025-03-02
Add Card:can_calculate

VictinZero:
Due to how vanilla calculation is implemented, it can be difficult to remove or otherwise manipulate Cards during the main scoring phase, and perhaps some other moments. This PR aims to introduce a `Card:can_calculate` method that should enable modders to verify that a card can be expected to exist. This usage is accomplished in the following way:

1. Call `card:can_calculate(true)` to decide if an effect should be applied to/from the `card` (we'll explain the argument below);
2. Set `card.getting_sliced = true` when creating an Event that will remove or otherwise prevent a card from being calculated or "targeted" by other effects.

In addition to the aforementioned utility, the method also tests if the `card` object is debuffed or not. Thus, it can be used more generally to test if the `card` "can be calculated", as the name suggests. The first and only argument of the method, when `true`, ignores the debuff status of the `card`.

The method also provides a patch target for mods that introduce mechanics that may prevent a card from being calculated, if `card.getting_sliced` is not enough.

**Commentary:** Personally, I think methods such as `Card:get_chip_bonus` shouldn't call `Card:can_calculate`. Instead, I think `Card:can_calculate` should be called to decide if `Card:get_chip_bonus` is called or not. This is because I see the objective of the method to provide the number, and for that number to be used for some purpose (_e.g._ for calculation). Since users can access the values directly, it might not be a big loss, but I think encouraging users to use methods leads to more compatible mods. (Compare the availability of `Card:is_suit` with the lack of `Card:is_rank`.)

Additionally, I didn't touch the vanilla Jokers' checks. I think there's some subjective arguments on whether or not they should be changed.

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -34,7 +34,7 @@ position = 'at'
 match_indent = true
 payload = '''
 function eval_card(card, context)
-    if card.ability.set ~= 'Joker' and card.debuff then return {}, {} end
+    if card.ability.set ~= 'Joker' and not card:can_calculate() then return {}, {} end
     context = context or {}
     local ret = {}
 '''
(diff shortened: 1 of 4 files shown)


VictinZero:
An extra comment that I forgot: some mods may want to add effects that trigger _e.g._ when a Card _itself_ is destroyed. Since that's not a vanilla mechanic, it might be good to give some thought how those effects could be implemented if this PR is merged.

(I do think a SMODS utility to handle that would be useful, but it's out of scope for this PR.)

PR #529, opened 2025-03-09
Info tabs as GameObject

baliame:
Adds SMODS.Tab GameObject. Does nothing on its own, but objectifies tab-based dialog rendering. Implements Run Info and Deck Info as tab groups, allowing mods to add their own tabs quickly.

## Additional Info:

### DIFF
--- a/lsp_def/classes/tab.lua
+++ b/lsp_def/classes/tab.lua
@@ -1,0 +1,43 @@
+---@meta
+
+---@class SMODS.Tab: SMODS.GameObject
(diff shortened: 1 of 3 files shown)


baliame:
This should be ready for review again, I'm not planning to add more UI classes until someone tells me if we even want this direction.

english5040:
Hello- looks like no one's reviewing your PR. I see your PR, it seems that people are just busy.

I definitely like the general idea. I'll take a deeper look when I have time

PR #548, opened 2025-03-15
Better calc 2.3

Aurelius7309:
## What's changed
- Structural changes to functions processing calculations
  - Reorganized `SMODS.calculate_context` using a helper function
  - "Flags" returned by `SMODS.trigger_effects`
  - `Card:can_calculate` is no longer called inside functions called from `eval_card`. `context.ignore_debuff` can be used to make debuffed cards be calculated in a context.
- Cards are no longer juiced up independently of the effect key in `SMODS.calculate_individual_effect` to account for new return values
- Added "individual objects" into the calculation pipeline. Deck effects have been integrated into this system. 
- Added the current blind as an individual object
- Added blind functions as calculation contexts: `blind_disabled`, `blind_defeated`, `press_play`, `debuff_card`, `debuff_hand`, `stay_flipped`, `modify_hand` 

## Additional Info:


moxie-coder:
will this cause any breaking changes to some mods?

Aurelius7309:
There's a few changes to function signatures and return values in here. Mods that hook the same functions will need to adjust for those

PR #566, opened 2025-03-20
added nil checks to blind calculations

Aikoyori:
This PR adds nil checks to the new blind calculations so current mods don't break when `flags` are nil

Tested with latest Cryptid. Should fix the crash at least for now if nothing else needed attention.

## Additional Info:

- [ / ] I didn't modify api's or I've made a PR to the [wiki repo](https://github.com/Steamodded/wiki).
- [ / ] I didn't modify api's or I've updated lsp definitions.
- [ / ] I didn't make new lovely files or all new lovely files have appropriate priority.


Aurelius7309:
I think this is unwanted, as now mods that hook these functions incorrectly would silently break the calculations using this instead of crashing straight-up. I also had the hook in question resolved on Cryptid's end before hand

PR #581, opened 2025-03-23
setting playing cards to negative only draws cards during blind

vitellaryjr:
prevents an issue where setting cards to negative during round end draws cards to the hand that doesn't exist
## Additional Info:


english5040:
Seems fine to me. If you tested it, I'll just take your word for it and merge, this is a tiny change.

PR #582, opened 2025-03-24
Deck Preview Pages

RattlingSnow353:
- [ ] I changed the overrides file, only the function G.UIDEF.view_deck while adding G.FUNCS.your_suits_page

Screenshot 2025-03-24 104721


RattlingSnow353:
> I believe the only problem I have with it right now is played cards aren't greyed

I fixed this

PR #593, opened 2025-03-28
add SMODS.Colour API

vitellaryjr:
SMODS.Gradient now extends it, shouldn't be any different in function

## Additional Info:

  - if this gets merged i'll, probably look into updating the SMODS.Gradient page to also include SMODS.Colour later today


nh6574:
I believe adding a color to the G.C table doesn't really need a whole API method dedicated to it. If you still want to, you can create an `SMODS.Gradient` with only 1 color to basically recreate this functionality.

PR #604, opened 2025-04-03
Blue Seal crash fix

SageJFox:
Prevents crash from Blue Seal attempting to create a non-existent planet card for a custom winning hand that lacks one. (Causes a random planet card to be rewarded instead.)

## Additional Info:

### DIFF
--- a/lovely/fixes.toml
+++ b/lovely/fixes.toml
@@ -691,3 +691,13 @@ payload = """if SMODS.seeing_double_check(context.scoring_hand, 'Clubs') then
 end"""
 overwrite = true
 match_indent = true
+
+# Card:get_end_of_round_effect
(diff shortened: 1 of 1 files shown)


english5040:
My instinct is that it wouldn't give a planet card at all. But this is bikeshedding I guess, we definitely need to do this for mod compatibility so I'm ok with it as is

SageJFox:
> My instinct is that it wouldn't give a planet card at all. But this is bikeshedding I guess, we definitely need to do this for mod compatibility so I'm ok with it as is

I'd be fine either way myself, as the not crashing the game is the important part to me. But the idea with the random planet is that it both at least makes the blue seal do *something*, and was its old behavior so I feel that it's not totally out there.

Could also split the difference and make it always give a Pluto, which'd just be making it 'c_pluto' instead of nil.

PR #605, opened 2025-04-03
Fix crash when area.cards is nil in SMODS.find_card

its-edalo:
In rare cases, the `cards` field of the `area` variable in `SMODS.find_card` can be nil, which causes a crash when calling `pairs(area.cards)`.

Crash example clip:
https://www.twitch.tv/edzyttv/clip/ColdAliveTruffleGingerPower-6tN0hLfSlknG8t_4

A simple nil check has been added to prevent the issue.

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -344,9 +344,11 @@ function SMODS.find_card(key, count_debuffed)
     local results = {}
     if not G.jokers or not G.jokers.cards then return {} end
     for _, area in ipairs(SMODS.get_card_areas('jokers')) do
-        for _, v in pairs(area.cards) do
-            if v and type(v) == 'table' and v.config.center.key == key and (count_debuffed or not v.debuff) then
-                table.insert(results, v)
+        if area.cards then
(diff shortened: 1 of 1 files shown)


english5040:
This seems like a small change. I'm not sure how a card area could possibly not have a `cards` field, though, so I'm wary of adding this check.

Have you found a specific mod that causes the problem?

its-edalo:
Admittedly I've never experienced the crash myself, but I've seen it happen several times when watching streamers - including the clip included. I think all of them were just switching between playing the multiplayer mod to single player.

My (uneducated) guess would be leftovers from how the multiplayer mod handles cards that your opponent played but need to show on your screen.

In any case, this does seem like a check that would be good to have - if an area doesn't have a `cards` field for some reason, in particular we don't want or need to search it for our particular card.

english5040:
Thanks for replying, that info about switching off from the multiplayer mod is really useful. I was thinking that this problem might be a symptom of some other issue. Anyway, I'll throw this PR in.

PR #607, opened 2025-04-04
SMODS.has_seal

TheRealNova422-real:
added a SMODS.has_seal function, it's SMODS.has_enhancement but seals 
(might also be nice to have a has_edition and has_sticker function too but idk how to do that)

## Additional Info:

- [V] I didn't make new lovely files or all new lovely files have appropriate priority.

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -974,6 +974,10 @@ function SMODS.has_enhancement(card, key)
     return false
 end
 
+function SMODS.has_seal(card, key)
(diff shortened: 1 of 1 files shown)


Aurelius7309:
What's the point of this existing? For it to be at all potentially useful, we'd have to enforce its use everywhere, which is a breaking API change. It seems to achieve nothing as it doesn't implement quantum seals

PR #611, opened 2025-04-05
Fix incorrect removal of used_jokers

random-facades:
When a card is changed into a new joker, the old joker is still set in `G.GAME.used_jokers`
This is due to the location of the `G.GAME.used_jokers[old_center.key] = nil` code.
It occurs before you change the current card's key, so there will still be a joker present to be found with `SMODS.find_card(old_center.key, true)`.

I shifted the code to be after changing the `self.config.center` which will remove it from any `find_card` search and will remove it from `used_jokers` correctly.

## Additional Info:

### DIFF
--- a/lovely/pool.toml
+++ b/lovely/pool.toml
@@ -142,7 +142,7 @@ end
 [[patches]]
 [patches.pattern]
 target = "card.lua"
-pattern = "local old_center = self.config.center"
+pattern = "self.config.center = center"
 position = 'after'
 payload = '''
 if old_center and not next(SMODS.find_card(old_center.key, true)) then


VictinZero:
As I mentioned on the Discord, my suspicion is that transforming a Joker simply by calling `Card:set_ability` is incorrect, thus I question the necessity of this PR.

Furthermore, another modder seems to have been unable to replicate this issue.

english5040:
Yes, you're not supposed to directly call set_ability to transform a Joker in vanilla. The current code was supposed to fix that, looks like it doesn't work.

This change seems reasonable but I definitely need to check if there are any logical issues.

VictinZero:
We talked about it over Discord. It seems that the user who couldn't reproduce it got tripped by DebugPlus' not handling `G.GAME.used_jokers`. I had the impression @Eremel would PR their implementation of transforming a Joker, but I'm not sure.

Simply using `Card:set_ability` is not sufficient as it does not handle some sprite scaling. For example: transforming Wee Joker into a non-Wee Card does not change its size. (The reverse transformation does change size as expected.) Thus it seems to me like an utility function implementing this functionality would be useful. It could maybe extend `copy_card` (like `SMODS.create_card` does to `create_card`).

(Just using `copy_card` to transform doesn't work, as I found out yesterday…)

Eremel:
I'll add my sprite size changes later today, juggling PRs on mobile is not fun. `Card:set_ability` should be the correct function to transform a card, there's no need for a utility function.

VictinZero:
> Card:set_ability should be the correct function to transform a card […]
What makes you say that?

When copying cards, I see that thunk wrote an entire function for it rather than simply setting the card's ability. I had originally assumed it was to properly handle sprites and copy table values, though in hindsight it seems to not do the former. Is the only reason `copy_card` exists to copy table values, or does it do something else that is necessary when transforming a card?

Aurelius7309:
The reason copy_card is used by thunk is because the card being transformed (e.g. by Death) should be an exact copy of the copied card in all aspects, not just the center. With set_ability, your goal is to preserve other properties of the card while only changing the center

PR #617, opened 2025-04-07
fix crash without wins_by_key

Mysthaps:
💀💀💀

### DIFF
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -1643,7 +1643,7 @@ function get_joker_win_sticker(_center, index)
 		local applied = {}
 		local _count = 0
 		local _stake = nil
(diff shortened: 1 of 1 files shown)


english5040:
Alright, this seems simple enough. Not merging immediately because the function convert_save_data should have prevented this from happening, so this issue might be due to some other bug.

Do you have any info on what caused this bug?

Mysthaps:
oops never checked my github again after this
this happens when save file is modified "incorrectly" by a mod, incorrectly as in without wins_by_key which is usually not needed to be added

english5040:
yeah makes sense mods could modify it wrongly. this is good

english5040:
did I not merge this? sorry

PR #626, opened 2025-04-11
Hand limit API

Eremel:
Adds `SMODS.change_play_limit(mod)` and `SMODS.change_discard_limit(mod)` to control the number of cards you are able to play and discard independently of each other.

Adding as a PR then I don't scuff up everyone's lovely patches that do this effect in their own ways, and people can be prepared to adjust when it's merged.

## Additional Info:

### DIFF
--- a/localization/en-us.lua
+++ b/localization/en-us.lua
@@ -218,6 +218,7 @@ return {
             b_deckskins_lc = 'Low Contrast Colors',
             b_deckskins_hc = 'High Contrast Colors',
             b_deckskins_def = 'Default Colors',
+            b_limit = 'Limit: '
 		},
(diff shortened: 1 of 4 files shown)


english5040:
I don't think hand select/discard select should be hard capped at a min of 1/0- hand size goes below 0, the game keeps track of the effective negative hand size. Effective values should probably be consistent with that system

english5040:
I guess I should explain myself better- I see people calling the change_ functions and feeling like the result is unexpected, especially if they get an error and the change "doesn't go through". It shouldn't be on the user to have to handle this type of error as it is a natural thing to happen, like handsize happening to go below 0.

PR #636, opened 2025-04-16
More Explicit Asserts

GauntletGames-2086:
Makes currently existing asserts provide explicit error indicating problems, rather than leaving them blank. Additionally considering updating older ones for style. 

## Additional Info:

### DIFF
--- a/libs/https/smods-https.lua
+++ b/libs/https/smods-https.lua
@@ -73,10 +73,10 @@ end
 local methods = {GET=true, HEAD=true, POST=true, PUT=true, DELETE=true, PATCH=true}
 
 local function checkAndHandleInput(url, options, skipUserAgent)
-	assert(type(url) == "string", "url must be a string")
+	assert(type(url) == "string", "Assertion failed: \"url\" must be a string")
 	options = options or {}
-	assert(type(options) == "table", "options must be a table")
-	assert(type(options.headers or {}) == "table", "options.headers must be a table")
+	assert(type(options) == "table", "Assertion failed: \"options\" must be a table")
+	assert(type(options.headers or {}) == "table", "Assertion failed: \"options.headers\" must be a table")
 	local contentTypeHeader = false
 	if not skipUserAgent then
 		local headers = {}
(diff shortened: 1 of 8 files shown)


english5040:
Hm, sometimes asserts indicate a specific problem with a mod or some function call did not satisfy its preconditions, but sometimes asserts indicate that some internal assumption in steamodded was violated and thus smods as a whole is bugged. Maybe clarify that?

eh. It's ok if we don't think too hard about this. Thanks for putting in the effort for code style, I approve

GauntletGames-2086:
> Hm, sometimes asserts indicate a specific problem with a mod or some function call did not satisfy its preconditions, but sometimes asserts indicate that some internal assumption in steamodded was violated and thus smods as a whole is bugged. Maybe clarify that?

Agreed, although none of the asserts (besides `o.mod == nil`, but someone actually ran into that one a few hours ago) are really internal as much as sanity checks you can realistically run into from human error. There's some save-related ones targeting `_saved_d_u` and `_discovered_unlocked_overwritten` that I couldn't really get a grasp on what they're trying to check for and omitted, which I think would be an example of an internal assert that could be expressed differently (as they look like they're meant for internal saving errors).

english5040:
> There's some save-related ones targeting `_saved_d_u` and `_discovered_unlocked_overwritten` that I couldn't really get a grasp on what they're trying to check for and omitted, which I think would be an example of an internal assert that could be expressed differently (as they look like they're meant for internal saving errors).

I think that was my code oops- the asserts should have to do with needing to save the old discovered/unlocked state, without saving an overwritten version and not overwriting multiple times, etc. I can take a look myself

english5040:
I'll go update the lsp and document the variables that are used in that bit of code, that will help

english5040:
This PR is a bit stale. Wilson already commented that they don't like printing "Assertion Failed" everywhere, fair.

How about marking internal errors with something like "Internal:" or "SMODS:", and for user errors just keep the part after "Assertion Failed"

GauntletGames-2086:
> This PR is a bit stale. Wilson already commented that they don't like printing "Assertion Failed" everywhere, fair.
> 
> How about marking internal errors with something like "Internal:" or "SMODS:", and for user errors just keep the part after "Assertion Failed"

Gotya, updating style now. Assuming that's it this should be clear to merge, as the rest of the PRs can just be retrofitted if they happen to include any blank asserts.

PR #646, opened 2025-04-21
Add rotation/offset support for shiny stickers in SMODS.DrawStep

larswijn:
Adds support for sticker rotation/offset for shiny stickers in SMODS.DrawStep by properly rotating/offsetting both the sticker itself and the "shiny" shader instead of only the sticker.

Also, this code was originally my badly implemented way to rotate/offset the sticker and can result in visual issues, so if there's a better method I don't know about let me know.

## Additional Info:


english5040:
Alright this is simple enough to immediately merge

As for your question, sorry, I'm probably not the one to ask

larswijn:
Thanks for the merge! And no worries about any better methods, it was a big shot in the dark.

Would it be possible to have an smods version bump with this?

PR #653, opened 2025-04-25
Better catch for non-vanilla scaling factors

wingedcatgirl:
Vanilla `get_blind_amount` expects exactly 1, 2, 3, or nil; call `SMODS.get_blind_amount` for _any_ other value rather than only values exceeding 3. (e.g. decimal, negative, zero)
## Additional Info:


english5040:
thank you, merging immediately

PR #655, opened 2025-04-26
Russian localization

EnderDrag0n:
## Additional Info:

### DIFF
--- a/localization/ru.lua
+++ b/localization/ru.lua
@@ -1,0 +1,237 @@
+return {
+    descriptions = {
+        Other = {
+            load_success = {
+                text = {
+                    'Мод загружен',
+                    '{C:green}успешно!'
+                }
+            },
+            load_failure_d = {
+                text = {
+                    'Не хватает {C:attention}зависимостей!',
+                    '#1#',
+                }
+            },
+            load_failure_c = {
+                text = {
+                    'Неразрешенные {C:attention}конфликты!',
+                    '#1#'
+                }
+            },
+            load_failure_d_c = {
+                text = {
+                    'Не хватает {C:attention}зависимостей!',
+                    '#1#',
+                    'Неразрешенные {C:attention}конфликты!',
+                    '#1#'
+                }
+            },
+            load_failure_o = {
+                text = {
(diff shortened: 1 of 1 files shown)


SpaD-Overolls:
"load_failure_d_c" is using a duplicate #1# instead of the intended #2# 
"load_failure_p" refers to "prefixes" and "mods" in plural, when it should be referring to them in singular

the stings for extra modifiers use "экстра" when a more appropriate translation would be "дополнительных" (which is admittedly too long, but it could be shortened to "доп.")

"card_extra_mult", "card_x_mult", "card_extra_x_mult", "card_extra_h_mult", "card_h_x_mult" and "card_extra_h_x_mult" refer to Mult as "Модификатор", when vanilla refers to it as "Множ."

"card_extra_p_dollars" is untranslated

"e_negative_playing_card" has its name in masculine, when it should probably be in feminine ([Негативный] in vanilla is for Джокер)

the enhancement descriptions are inconsistent with vanilla's descriptions (SMODS includes these entries to remove the + and $, other than the lack of those symbols they should be the same as vanilla)

"b_credits" should be "Авторы", as "Титры" means a specific kind of credits (staff roll), a descriptor not applicable to the word's usage in SMODS

"b_rewrite_by" should be changed to something else, as "переписан" means "copied over" (maybe "переработка от "?)

"b_github_project" has incorrect attributive order for russian (proper form would be "Проект Github", although contextually "Наш Github" would make more sense to a russian reader)

"b_github_bugs_1" and "b_github_bugs_2" have incorrect theme-rheme order (my suggestion for this would be "Там Вы можете сообщить нам" "о багах и помочь в разработке")

"b_lovely_mod" is untranslated ("Мод на Lovely" would be the best translation that is similar in length)

for "b_additions" "Нововведения" would fit better contextually (that button is for showing all the things added by a mod)

"b_seeded_unlocks_info" and "smods_version_mismatch" refer to runs as "забег" when vanilla uses "партия" instead

strings for "ml_achievement_settings" have typos (there's no ё in plural "Включены" and "Отключены", as the stress falls onto the last syllable)

english5040:
Note to self to merge

PR #659, opened 2025-04-29
DeckSkin Card UI (Crediting)

Kekulism:
## Changes:
- Two functions have been added to the SMODS.DeckSkin object _(with names subject to change)_, **generate_ds_card_ui**, and **has_ds_card_ui**. Both of these functions are run for each card that is displayed when previewing their respective DeckSkin, and are run again every time the palette is changed as well. First, each card runs has_ds_card_ui. If the return of the function is true, this card will now have its no_ui value turned off, allowing it to be hovered over and see its ui in the Customize Deck menu. Next, since the card has a ui, when its hovered over, the card will run generate_ds_card_ui to overwrite its default ui (showing its chip value) with whatever code is ran in the function.
- Added code that automatically adds credits to any Friends of Jimbo collabs that are credited to a specific artist in the game's Collabs tab in the credits. During the game's launch, the code will automatically initialize the G.collab_credits table by forcing the menu to open, then immediately closing it, which is never visually seen in game as it happens before the splash screen. For every Friends of Jimbo Collab, the two new functions will check if that deckskin has a match in G.collab_credits, and if it does, it will automatically add the artist's credit to the card ui.

Discord_khBfjqwEs8

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@ Click [here](https://github.com/Steamopollys/Steamodded/wiki).
 - Put the mod into that directory. (The mod can be a single file if there is only one file provided, or it can be a whole folder.)
 - Launch the game and enjoy!
 
+## Creating a Mod
+
+Useful resources for creating your first mod can be found [here](https://github.com/Steamodded/smods/wiki/Your-First-Mod).
+
 ## Features
 
 Documentation for this project is currently incomplete. A collection of documentation pages and guides that are currently available can be found [here](https://github.com/Steamopollys/Steamodded/wiki).
--- a/localization/en-us.lua
+++ b/localization/en-us.lua
@@ -136,6 +136,17 @@ return {
                     "{C:money}#1#{} if held at end of round",
                 },
             },
+            artist = {
+                text = {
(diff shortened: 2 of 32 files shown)


Kekulism:
OK i didnt realize that making new commits to my fork wouldn't make a new PR for this so I guess these new tweaks are just here now. Lemme explain em:

- Added small functions for the Straight/Flush modifying Jokers. Rather than being hardcoded into the functions, Straights and Flushes now use SMODS.shortcut(), SMODS.four_fingers(), and SMODS.wrap_around_straight() to modify the hands. By default, SMODS.shortcut() will return true if it finds Shortcut,  SMODS.four_fingers() will return 4 if it finds 4 fingers but return 5 if it doesn't, and SMODS.wrap_around_straight() will always return false. Modders can now use function patching to patch into these functions to easily utilize them without having to lovely patch.
- Added new trigger, blocking and blockable variables to extra in card_eval_status_text to allow for changing the event timing.
- Joker Sticker counter will now check for if a joker is not omitted before tracking its stickers.

PR #661, opened 2025-04-29
Dedicated functions for listed probabilities

TamerSoup625:
Some functions have been added to SMODS:

- `SMODS.get_probability_vars(trigger_obj, base_numerator, base_denominator)`: Returns the numerator and denominator of a listed probability with appropriate modifications.
- `SMODS.pseudorandom_probability(trigger_obj, seed, base_numerator, base_denominator)`: Runs a listed probability check. Basically equal to `pseudorandom(seed) < G.GAME.probabilities.normal*base_numerator/base_denominator`.
- `SMODS.get_probability_vars_table(trigger_obj, base_numerator, base_denominator)`: An alternate version of `SMODS.get_probability_vars` that returns a tuple instead of two numbers.

This PR also includes LSP definitions and all the patches and overrides needed such that these methods are used in vanilla.

This suggestions makes working with listed probabilities generally much easier. Not only the pseudorandom comparisions are made by a function instead of manually, but you can hook `SMODS.get_probability_vars` for greater control of listed probabilities (for example +probabilities or fixed probabilities) or hook `SMODS.pseudorandom_probability` to run code when a listed probability succeeds or fails.

## Additional Info:

### DIFF
--- a/localization/en-us.lua
+++ b/localization/en-us.lua
@@ -167,6 +167,15 @@ return {
                     "{C:mult}#1#{} Mult",
                 },
             },
+            m_lucky={
+                name="Lucky Card",
+                text={
+                    "{C:green}#1# in #3#{} chance",
+                    "for {C:mult}+#2#{} Mult",
+                    "{C:green}#6# in #5#{} chance",
+                    "to win {C:money}$#4#",
+                },
+            },
         }
     },
     misc = {
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -603,7 +603,7 @@ for j = 1, #G.jokers.cards do
 
 end
 
-if SMODS.has_enhancement(scoring_hand[i], 'm_glass') and scoring_hand[i]:can_calculate() and pseudorandom('glass') < G.GAME.probabilities.normal/(scoring_hand[i].ability.name == 'Glass Card' and scoring_hand[i].ability.extra or G.P_CENTERS.m_glass.config.extra) then
+if SMODS.has_enhancement(scoring_hand[i], 'm_glass') and scoring_hand[i]:can_calculate() and SMODS.pseudorandom_probability(scoring_hand[i], 'glass', 1, scoring_hand[i].ability.name == 'Glass Card' and scoring_hand[i].ability.extra or G.P_CENTERS.m_glass.config.extra) then
     destroyed = true
 end
 
--- a/lovely/listed_probabilities.toml
+++ b/lovely/listed_probabilities.toml
@@ -1,0 +1,52 @@
+[manifest]
+version = "1.0.0"
+dump_lua = true
+priority = -5
+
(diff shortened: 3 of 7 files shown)


Eremel (commented):
I like the general idea of this PR, but I would worry about recommending hooking to modify the values as the primary method. I think adding some target patch lines, *(similar to how a lot of SMODS calculation functions have)*, for additive modifiers, multiplicative modifiers, and fixing the number modifiers would be preferable.

I'm not 100% sure on the necessity of passing the triggering object *(I'm assuming it's for effects that modify probabilities of specific objects)* but I think I would reorder the variables so they go  `numerator, denominator, seed, trigger_obj` and have some default values for seeds and trigger_obj so the function can just be used as `SMODS.pseudorandom_probability(1, 4)` for example.

The `SMODS.get_probability_vars_table` function seems unnecessary, particularly as you can use `{SMODS.get_probability_vars()}` to get the values in a table form.

There's also the issue of getting the community to take up using this, though I'm not sure how to approach that just yet.

english5040:
yeah, these issues are very fixable

TamerSoup625:
I made `SMODS.pseudorandom_probability` take `seed` and `trigger_obj` first such that you wouldn't ignore them (`seed` ensures consistency with seeded runs, and `trigger_obj` can modify probabilities based on which card it's queued), similar to how `level_up_hand` has `card` as its first argument.

Thinking back about it, I also find using hooks not ideal for modifying probabilities since there's the problem of order of operations, so I'll definitely touch up this part when I have the chance. Using target patch lines could work, but I've also heard someone suggest using a calculation context. In my opinion, the best approach would be allowing both of these methods for a finer control.

Eremel:
I'm fine with having `trigger_obj` before the values in the function to encourage use of it, but I think `seed` should definitely be the last argument, and add support for a default value for a "general probability" seeded queue.

Yeah, I was considering some contexts the other day briefly. I think a `{add_probability = true, numerator = X, denominator = Y}`, a `{scale_probability = true, numerator = X, denominator = Y}` and a `{fix_probability = true, numerator = X, denominator = Y}` set up would work best, with some collection for `numerator` and `denominator` to handle it automatically. These would work better than target patch lines now that calculation is consistent across almost all objects.

I'll take a look at implementing the context calculation later today.

TamerSoup625:
I'd still go in favor of `seed` as the first/second argument, but it could be instead optional (`string|nil`) and if you pass `nil` it'll do an unseeded probability queue.

In my opinion, the calculation context should be unified into `{mod_probability = true}`. To modify probability, we can either have the numerator and denominator as read and write context values (+ `trigger_obj` as read-only), or have the calculate function return a function with the signature of `SMODS.get_probability_vars`, after which all functions returned from calculation will be called to get the final fraction. This way, probability will be modified following calculation order, and in case you want to follow another order you can use patch lines.

PR #672, opened 2025-05-04
Add enhancements to poker hand tips

nh6574:
Added as a third value to the examples

Captura de pantalla 2025-05-04 094403

## Additional Info:


english5040:
On first glance I don't like the implementation- it's not general enough and we love generality in APIs.
The general idea is good

BakersDozenBagels:
You could always require something like `{ 'H_K', true, enhancement = 'm_gold' }`. With that, adding keys for other modifications isn't horrifically clunky.

nh6574:
Makes sense to me. I can add that if there's no second opinion

nh6574:
Changed it to @BakersDozenBagels's suggestion and added editions and seals as well. Could probably be expanded to stickers too.

image
Captura de pantalla 2025-05-06 211815

english5040:
Alright, this looks good

PR #675, opened 2025-05-05
Add context.hand_space

nh6574:
Added context.hand_space to recreate effects like The Serpent. `context.hand_space` has the amount of cards about to be drawn (after serpent calculates). `hand_space = integer` can be returned in that context to change the number of cards drawn.

Maybe it could have a more descriptive name but I couldn't come up with one :P

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -1001,7 +1001,7 @@ end
 payload = '''
 local flags = SMODS.calculate_context({hand_space = hand_space})
 print(flags)
-hand_space = flags.hand_space or hand_space
+hand_space = math.min(#G.deck.cards, flags.hand_space or hand_space)
 '''
 
 # Used to identify first hand of round
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -130,7 +130,7 @@ if context.cardarea == G.hand then
     end
 
     local jokers = card:calculate_joker(context)
-    if jokers then 
+    if jokers then
         ret.jokers = jokers
     end
 '''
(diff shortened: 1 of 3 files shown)


english5040:
I don't think this wants to be a calculation step- it seems like mods would want to insert actual code there, as the Serpent does, and might want to do other things with how cards are drawn. Also, it feels weird to have to read in a variable from context, and then set that variable through returning a flag- I'd like to just be able to change that variable directly

nh6574:
What do you mean insert code like Serpent? Serpent only modifies hand_space and nothing else. Examples I've seen from other mods only include manipulating hand_space and doing stuff before the cards are drawn, both of which this context would allow. Anything else could still be done by patching the later part that draws cards or hooking draw_card but I don't see how that impedes this being added as it wouldn't conflict with that (not more than Serpent and other patches to this part of the code would)

In my opinion, after the blind calculations were added it makes sense to be able to recreate vanilla blinds without having to patch manually.

english5040:
Actually I forgot about new blind calc; I do think hand_space probably doesn't want to be inputted and outputted in such an awkward way but I'm dropping all other objections. This is good for new blind calc

english5040:
ok, I'm realizing why this PR rubs me the wrong way: in https://github.com/Steamodded/smods/pull/548 there's a list of blind functions `blind_disabled`, etc. which were given calculation contexts. But they all have to do either with "doing something" at that point like `blind_disabled` or "deciding whether to do" a particular thing, like `debuff_hand`. The thing you're deciding whether to do is clear. With this timing window, drawing cards to hand is changing "how" you're doing something which is inherently awkward using a calculation function.

Comparing this to discarding cards: Trading Card can give you money and destroy a card on discard, but a joker that changed _how_ you discard would be tough. It's not something that _should_ be accomplished by passing stuff inside context and returning flags.

nh6574:
I completely agree that this is different than other contexts and the manipulation might feel awkward. I still disagree that this shouldn't be something done with a context and returning flags. 

Other than avoiding patches one thing this would solve is add predictability to the order of effects (for example, jokers to the right having last say on the effect) instead of relying on code order.

Do you think adding `SMODS.change_hand_space(mod)` and changing this context to something like context.draw_from_deck_to_hand would solve your issue? My problem with that would be making hand_space changes permanent, but maybe there's something else similar that could be done.

PR #677, opened 2025-05-05
Add support for negative chips/mult

chaseoqueso:
Added a hook for scale_number that accounts for the negative sign when determining the scale for a number
Added patches for the chip and mult update functions that prevent domain errors on the math.log10 calls by taking the absolute value

## Additional Info:


english5040:
This is just good for compatibility, I agree we need this. I think the code is a bit scuffed, especially the math in the hook; I'll edit the PR later

english5040:
Hm, there's a check in the game that means negative chips/mult will actually be displayed as an empty string. Closing this; negative numbers should be handled, but not in this way

PR #684, opened 2025-05-06
Add more text formatting options

Aikoyori:
## Summary
- This PR adds a way to define font using Steamodded API (can be used for defining a language or for use with text formatting)
- This PR also makes descriptions and name able to contain custom fonts 

Usage: "{font:internal font number or font key with prefix}text with font"

## Synopsis
Normally, Balatro does not support multiple fonts in tooltips. Which makes multilingual description less than trivial. With this PR, developers can use the default (and possibly standardized) `G.FONTS` indices or define their own fonts using `SMODS.Font` to further customise an item's name or description. This `SMODS.Font` can also be used to define localizations' font for more advanced parameters if so choose.

## Example
This jumble of text using multiple languages and formatting rules.
image
can be defined by this
image
image
Note that font:5 refers to the game's internal font for different localization. In this case, 5 refers to Noto JP font, the default Japanese localization font for Balatro.

Feel free to edit this PR if something was unintentionally broken.

## Additional Info:

### DIFF
--- a/lovely/ui_additional_text_props.toml
+++ b/lovely/ui_additional_text_props.toml
@@ -1,0 +1,211 @@
+[manifest]
+version = "1.2"
+dump_lua = true
+priority = 0
+
+[[patches]]
+[patches.pattern]
+target = "functions/misc_functions.lua"
+pattern = """              spacing = math.max(0, 0.32*(17 - #assembled_string)),"""
+position = "after"
+payload = """font = SMODS.Fonts[part.control.font] and SMODS.Fonts[part.control.font] or (tonumber(part.control.font) and G.FONTS[tonumber(part.control.font)]),"""
+overwrite = true
+match_indent = true
+
+[[patches]]
+[patches.pattern]
+target = "functions/misc_functions.lua"
+pattern = """            spacing = _spacing,"""
+position = "after"
+payload = """font = SMODS.Fonts[part.control.font] and SMODS.Fonts[part.control.font] or (tonumber(part.control.font) and G.FONTS[tonumber(part.control.font)]),"""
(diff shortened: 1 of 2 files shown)


english5040:
Trying to merge this. Docs are a bit inaccurate with what the parameters actually mean. Looking at the LOVE docs for fonts.

The lovely patch files seem really long, is it possible to simplify those?

english5040:
PR needs more time to be merged. Documentation needs to be accurate first:
- this only works for TrueType fonts according to LOVE's docs
- docs are unclear on the actual intent of render\_scale and FONTSCALE. I'm not sure what their meaning or intent is and if they're different, even if I know that what they do is scale. I think render\_scale multiplied by some stuff ends up as the parameter for the created font's size in pixels? (according to LOVE's docs)
- How do I determine TEXT_OFFSET, is it just trial and error? seems error prone
Overall: I am not sure how to use the API besides fiddling with the numbers through trial and error. Ideally using a font should "just work", maybe some of these values can be automatically set looking at LOVE's Font class?

I have not yet looked into simplifying the lovely patches and there may be additional issues in the documentation because I haven't looked at LOVE's docs in detail.

Help would be appreciated on these points. I realize the font code in Balatro is complicated and annoying so improving the docs might not be feasible, maybe all we can do is tell people to use trial and error

Aikoyori:
I agree on the documentation part. The font documentation is pretty hard to do as there's not much prior documentation about this and I'm just going from my understanding of the code.

Eremel (commented):
Made a couple of changes here, if you could check it all still works fine on your end I'm happy to merge and add it to the list of things needing docs 🙃

PR #687, opened 2025-05-08
fix `restart_game` referencing wrong path

nnmrts:
Ideally this should be determined dynamically, but this works as a quick fix, I guess.


english5040:
Yeah this can unfortunately only be hardcoded really

PR #690, opened 2025-05-08
fix hc art for custom rank + custom suit cards

MetaNite64:
If a mod adds art for a custom rank in a custom suit, and that custom suit is set to high-contrast art and the default deck skin, the custom ranks would still load the low-contrast art. This fixes that.

## Additional Info:


english5040:
thanks, saw your post in Paperback

PR #705, opened 2025-05-18
Fix base HC setting

notTamion:
makes the base game HC setting work again. Decided to not update the internal `colourblind_option` when someone updates HC using the Customization menu as the `colourblind_option` is used in a couple other places other than just cards, which will still let people turn that on/off independently.

### DIFF
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -1832,6 +1832,27 @@ G.FUNCS.change_collab = function(args)
 	G:save_settings()
 end
 
+G.FUNCS.refresh_contrast_mode = function ()
+	for i, suit in ipairs(SMODS.Suit:obj_list(true)) do
+		if G.COLLABS.options[suit.key] then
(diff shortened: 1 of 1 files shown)


Eremel (commented):
I'm not 100% sure what this is actually doing.  It seems that it overrides the individual control of high contrast colours by using the setting in the options instead?

Eremel (commented):
I'm not convinced that toggling that setting should modify the individual customisations of the deck skins themselves. Perhaps the button in settings should be removed/point to the deckskins menu, unless it controls other things that I'm not aware of?

notTamion:
> unless it controls other things that I'm not aware of?

i believe it does control some icons aswell, unless that was moved to the `customize deck` menu aswell.

I would personally prefer having it update all suits as that way its consistent with the base game and its just a nice qol thing so you don't have to go and change it manually for every suit. Though i am also fine with adding a redirect to the menu instead. Just right now it does basically nothing so its pretty much just randomly sitting there

PR #717, opened 2025-05-26
`SMODS.showman()`

nh6574:
Adds a function for mods to hook when the game checks for Showman to allow Showman-like effects. Additionally, it takes the key of the card to generate in order to be more granular with the effect.

Following #659 I've noticed updating [Vanilla Remade](https://github.com/nh6574/VanillaRemade) that Showman was the only joker to not have a graceful implementation using only SMODS and hooks (though the solution was simply to force the name to be "Showman")

## Additional Info:

### DIFF
--- a/lovely/center.toml
+++ b/lovely/center.toml
@@ -367,7 +367,7 @@ match_indent = true
 position = 'after'
 payload = '''
     for _, v in ipairs(SMODS.Consumable.legendaries) do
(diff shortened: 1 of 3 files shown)


english5040:
this looks good. I am actually going to raise an issue with the naming of these functions; they're not really about checking if four_fingers or whatever is present, sometimes it's "how long can a straight be".

this shouldn't block merging of course. I both love and hate API design, naming is utterly critical to APIs but being careful like this is just too slow

PR #719, opened 2025-05-28
`get_current_pool` for editions

nh6574:
Added the ability to use `get_current_pool("Edition")` and modified `poll_edition` to use it. Also added `in_shop = true` to the vanilla editions for this change.

## Additional Info:

### DIFF
--- a/lovely/pool.toml
+++ b/lovely/pool.toml
@@ -119,6 +119,20 @@ pool_opts = pool_opts or {}
 [[patches]]
 [patches.pattern]
 target = 'functions/common_events.lua'
+pattern = '''
+if _type == 'Enhanced' then
+    add = true
+'''
+match_indent = true
+position = 'after'
+payload = '''
+elseif _type == 'Edition' then
+    if v.in_shop then add = true end
+'''
+
+[[patches]]
+[patches.pattern]
+target = 'functions/common_events.lua'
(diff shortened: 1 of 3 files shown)


english5040:
Making get_current_pool more flexible is definitely good. There is a bit of awkwardness for the case of Editions: should "base edition" be an option basically.

Future concerns:
We have a bit of a consistency issue with some pools caring about weight and most not. Editions also having a "base edition" doesn't help.

nh6574:
> Future concerns: We have a bit of a consistency issue with some pools caring about weight and most not. Editions also having a "base edition" doesn't help.

I'm probably not doing it myself but IcyEthics is working on modifying my per-card rarity system (basically giving each card a weight) to make it more generic and hopefully PR it to SMODS. Maybe that will help with this issue. Or make it worse haha

Eremel:
I'm not sure that `get_current_pool` should consider weights, they should be considered in the various poll functions and the selection from the pool. `get_current_pool` should simply return the items that are available to be selected.

english5040:
is this to do with the current PR? sorry, I meant that overall, weights and random selection functions are confusing, not that get_current_pool should consider weights.

Eremel:
Looks good overall. Will merge in next wave.

I've started some work on addressing weights and random selection functions, will throw it in a PR when it's finished.

PR #720, opened 2025-05-28
(Better) Playing card support for SMODS.create_card

nh6574:
Better handling of playing cards with SMODS.create_card and add_card. 

`set` now takes `Playing Card` for a random choice between `Base` and `Enhanced` and all three sets now default to `G.hand` as the `area` instead of `G.jokers`.

Additionally the following parameters are added:

`front`: Front of the playing card, formatted like `S_R`. Added to maintain a formatting similar to `create_playing_card` for those who need it. Ignores `rank` and `suit`.
`rank`: Takes both the `key` or the `card_key` (allows integers for vanilla ranks).
`suit`: Takes both the `key` or the `card_key`
`enhanced_poll`: To modify the chance of choosing `Base` or `Enhanced` with set `Playing Card`. Default: 0.6

`SMODS.add_card` adds to the `G.playing_card` counter and inserts the card into `G.playing_cards`. I thought it made more sense than `SMODS.create_card` doing it.

## Additional Info:

### DIFF
--- a/lovely/center.toml
+++ b/lovely/center.toml
@@ -377,6 +377,15 @@ payload = '''
 [[patches]]
 [patches.pattern]
 target = 'functions/common_events.lua'
+pattern = "local card = Card(area.T.x + area.T.w/2, area.T.y, G.CARD_W, G.CARD_H, front, center,"
+match_indent = true
+position = 'at'
+payload = '''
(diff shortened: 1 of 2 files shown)


english5040:
I think having an easier time using create_card is great

Eremel:
Will merge in next wave.

PR #721, opened 2025-05-28
Add dollar_buffer calculation to `dollars` return (Resolves #709)

### LINKED ISSUE #709: Request: Update `G.GAME.dollar_buffer` automatically when returning `dollars`
This is something most mods don't do and from what I can tell it causes issues with Jokers like Bootstraps that get the dollar value during scoring. It would probably cause small issues with the mods that actually do it, however.

I raise this issue because I see people copying parts of my solution from VanillaRemade without understanding what it means and I feel like it might be more straightforward if it was handled by SMODS directly.

nh6574:
Adds to the dollar_buffer automatically when returning `dollars` in calculations.

I haven't seen this done in mods often or recommended to people as it's somewhat cumbersome to do, but it's important for cards to work properly with other cards that use the dollar amount during scoring.

It will probably mess with mods that are already doing it manually, however.

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -1250,6 +1250,7 @@ SMODS.calculate_individual_effect = function(effect, scored_card, key, amount, f
 
     if (key == 'p_dollars' or key == 'dollars' or key == 'h_dollars') and amount then
         if effect.card and effect.card ~= scored_card then juice_card(effect.card) end
+        G.GAME.dollar_buffer = (G.GAME.dollar_buffer or 0) + amount
         ease_dollars(amount)
         if not effect.remove_default_message then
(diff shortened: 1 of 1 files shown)


english5040:
the fact that it messes with mods that already do it is a serious problem, yeah. many mods make an error with this, but just as many are already doing it right.

english5040:
I can't think of a way to resolve this right now. it's a shame, this really trips a lot of mods up

nh6574:
Closing this as I want to clean up my repository and the PR is stale, not very important and easy to recreate. Leaving the issue open in case someone comes up with a better solution.

PR #723, opened 2025-05-28
Multiline name support and scaling fix

nh6574:
Adds support for multiline names and fixes the scaling when using text formatting.

image
Captura de pantalla 2025-05-28 163640

## Additional Info:

### DIFF
--- a/lovely/ui_additional_text_props.toml
+++ b/lovely/ui_additional_text_props.toml
@@ -140,6 +140,28 @@ end) then
 [patches.pattern]
 target = "functions/misc_functions.lua"
 pattern = '''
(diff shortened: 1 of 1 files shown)


nh6574:
I would love for people to do testing on this to see if there's anything this breaks or if there's any weird formatting issues I didn't notice.

Eremel (approved):
Seems to work in all instances I can think of that a name would be provided.

Will break mods that expect the name of a card to be a string rather than a table.

nh6574:
`name` is always a table, `name_text` is the string. So it shouldn't break anything about that.

Unless you mean they look directly at the localization but that already would be bad form I think

PR #726, opened 2025-05-31
context.round_eval

nh6574:
It adds the Anaglyph context to the regular calculation pipeline.

## Additional Info:


english5040:
yeah having contexts for things that are already happening is a good idea generally; hopefully deck calculation gets more integrated more into smods calc

Eremel:
Surprised this was missed when we adding deck calc, will merge in next wave.

PR #734, opened 2025-06-02
fix missing newlines

crazybot27:
## Additional Info:


english5040:
These newlines actually were critical in version 0.6 of Lovely: the output would be formatted incorrectly if the newline was there. I think this isn't worth worrying about, closing probably.

Thank you for the PR, sorry about this

PR #735, opened 2025-06-02
Add lookup table of ObjectTypes to each center

nh6574:
Adds an `object_types` field when a Center is inserted into an ObjectType to easily look up what types it belongs to.

## Additional Info:

### DIFF
--- a/src/game_object.lua
+++ b/src/game_object.lua
@@ -935,8 +935,8 @@ Set `prefix_config.key = false` on your object instead.]]):format(obj.key), obj.
         end,
         inject_card = function(self, center)
             if center.set ~= self.key then SMODS.insert_pool(G.P_CENTER_POOLS[self.key], center) end
-            if not center.object_types then center.object_types = {} end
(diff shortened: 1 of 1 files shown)


Eremel:
This would make checking objects are part of a pool much simpler. Will merge in next wave.

GauntletGames-2086:
Do we really need this to be a new parameter? `pools` is formatted identically to `object_types`, and already has the meaning of "add this card to X ObjectType" so it could make more sense to just update that field rather than making a whole new one.

nh6574:
I agree but I didn't know if doing that had any side effects. I can change it if it doesn't.

GauntletGames-2086:
> I agree but I didn't know if doing that had any side effects. I can change it if it doesn't.

I believe updates to GameObject tables get reset every time they get injected,which is the only moment where SMODS looks at the `pools` table. So no, there shouldn't be a side effect to updating that table during injection.

nh6574:
Done

PR #736, opened 2025-06-03
Add util function for merging calculate return tables

nh6574:
Adds `SMODS.merge_effects` that takes any number of arrays of tables and nests them using the `extra` field. 

Useful primarily for copying multiple cards with `SMODS.blueprint_effect`.

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -2442,3 +2442,40 @@ end
 function SMODS.wrap_around_straight()
     return false
 end
+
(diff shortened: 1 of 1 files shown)


english5040:
is this correct when one of the jokers copied manually executes its effect inside calculate()?

english5040:
I like the idea of this function. ok, maybe a bit niche. seems useful though

Issue #711, opened 2025-05-21
Better calc's animations not consistent with vanilla for Wee Joker


english5040:
In vanilla, Wee Joker moves when it upgrades. In better calc both Wee Joker and the scored 2 both move when Wee Joker upgrades.

nh6574:
This appears to be have been fixed but please reopen if I'm wrong

PR #737, opened 2025-06-03
Better calc individual juice fix (Issue #711)

Gloompashcy:
Slight change to the SMODS.calculate_effect function to make some cards stop juicing wrong

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -1472,7 +1472,14 @@ SMODS.calculate_effect = function(effect, scored_card, from_edition, pre_jokers)
     local ret = {}
     for _, key in ipairs(SMODS.calculation_keys) do
         if effect[key] then
-            if effect.juice_card then G.E_MANAGER:add_event(Event({trigger = 'immediate', func = function () effect.juice_card:juice_up(0.1); scored_card:juice_up(0.1); return true end})) end
+            if effect.juice_card then
+                G.E_MANAGER:add_event(Event({trigger = 'immediate', func = function ()
+                    effect.juice_card:juice_up(0.1)
+                    if (not effect.message_card) or (effect.message_card and effect.message_card ~= scored_card) then
+                        scored_card:juice_up(0.1)
+                    end
+                    return true end}))
+            end
             local calc = SMODS.calculate_individual_effect(effect, scored_card, key, effect[key], from_edition)
             if calc == true then ret.calculated = true end
             if type(calc) == 'string' then


english5040:
hey! thank you for making a PR!
for context: I made the original issue, thinking it's definitely Steamodded's fault. I'm generally suspicious of this animation bug, it might have a deeper cause- it might be that steamodded's better calc features just don't correctly assign card to some bog-standard jokers.

in short: fixing just Wee might be missing the root cause. I really appreciate the work though, finding what made wee wrong in the first place is really good.

english5040:
(just noticed your PR got closed. Github PRs require you to keep the branch not deleted, sorry)

Gloompashcy:
Yeah, I realized that almost all the cards that upgrade after a card is scored have this same issue, pretty sure it's because better calc always juices the card in play if a scored_card value isn't in the return table. I'll probably reopen with an fix for all of them

Gloompashcy:
Reopened with fixes for Lucky Cat and 8-Ball as well

Gloompashcy:
Temp fix for now, looking into a fix for the deeper issue

Eremel:
I'm not able to test right now, but I'm pretty sure that moving `effect.focus` to the start of the `or` chain in this [line](https://github.com/Steamodded/smods/blob/26625528a10ad579e229a56a98b1b28c8a5d6a16/src/utils.lua#L1307) would solve the problem without having to patch vanilla jokers. If you could test that for me, I'd appreciate that.

Gloompashcy:
That does fix this issue, but it also causes jokers like hiker to have scoring animations on the joker rather than the card

Gloompashcy:
May have found a working solution, testing now

Gloompashcy:
Seems to be working, but it is possible that an anim could break if it relies on two different cards juicing from effect and extra

Eremel:
Yeah looks like this would definitely break any jokers that are using the extra table to add messages to different things. I'll have a proper look after work later on and see what we can figure out.

Gloompashcy:
This is probably the most reasonable solution I've found currently. Since card_eval_status_text will juice the card, I've changed the scored_card:juice_up in SMODS.calculate_effect so that it only happens when the message card is not the same as scored_card

Gloompashcy:
This fixes the problem with wee, lucky cat and 8 ball, and hopefully shouldn't have an impact on uses of the extra table

PR #741, opened 2025-06-03
Allow `pseudorandom_element` to take strings for seeds

nh6574:
I'm tired of this one :)
Even the LSP says it does when it doesn't lol

## Additional Info:

### DIFF
--- a/lovely/pool.toml
+++ b/lovely/pool.toml
@@ -13,6 +13,7 @@ target = "functions/misc_functions.lua"
 pattern = "function pseudorandom_element(_t, seed)"
 position = "at"
 payload = """function pseudorandom_element(_t, seed, args)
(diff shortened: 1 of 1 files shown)


GauntletGames-2086:
If it's a problem with the LSP definition not being correct, then it is better to fix the LSP definition rather than retroactively changing the function, especially if this doesn't have a strong reason for it. 

As an API design decision, allowing the `seed` param to take a string just lets you move a required function call to `pseudorandom_element` rather than the callee. I don't think this is a significant enough problem to warrant a change like this.

nh6574:
I think it makes it inconsistent with `pseudorandom` which does take a string.

Edit: Oh, and inconsistent with #661 if that's implemented.

english5040:
This one's a good idea in my opinion. small change to reduce friction

PR #745, opened 2025-06-04
Added SMODS.JimboQuip

Kekulism:
Added a new GameObject, SMODS.JimboQuip which serves to allow mods to easily add new quips for Jimbo to say on run wins/losses. All you need is a key and a type set to 'win' or 'loss', as well as a loc_txt table or matching entry in a localization file.

Quips can be filtered out or increase their chance of appearing with the filter function. Additionally, you can set an extra table (or function that returns a table) to specify extra details like changing the center of the card character or the colors of its materialization or particles.

Decks can also filter out quips with an added quip_filter function to SMODS.Back. This allows certain decks to only show certain quips. quip_filter is called at the same time as filter is for every quip that can be added to the pool, and functions partially the same. returning false will prevent it from being shown.

**Required parameters:** `key`, `type`
**Basic Example:**
```
SMODS.JimboQuip{
    key = 'custom_wq_1,
    type = 'loss',
    loc_txt = {
        ['en-us'] = {
            "get gud",
        }
    }
}
```

**Advanced example:**
_This quip only appears if you win the run with Hanging Chad, and sets it rarity higher so you are almost guaranteed to see it. It also sets the center for the card character to appear as hanging chad instead of the basic joker, and changes its particle colors._
```
SMODS.JimboQuip{
    key = 'custom_wq_1',
    type = 'win',
    extra = { center = G.P_CENTERS.j_hanging_chad, particle_colours = { G.C.RED, G.C.WHITE, G.C.BLACK } }
    filter = function()
        if next(SMODS.find_card('j_hanging_chad')) then
            return true, { rarity = 100 }
        end
        return false
    end,
    loc_txt = {
        ['en-us'] = {
            "Hanging Chad is for Chads!",
        }
    }
}
```

## Additional Info:

- [X ] I didn't make new lovely files or all new lovely files have appropriate priority.

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -591,45 +591,6 @@ SMODS.calculate_context({full_hand = G.play.cards, scoring_hand = scoring_hand, 
 -- TARGET: effects before deck final_scoring_step
 '''
 
-# context.destroying_card
-[[patches]]
-[patches.pattern]
-target = "functions/state_events.lua"
-pattern = '''
-for j = 1, #G.jokers.cards do
-    destroyed = G.jokers.cards[j]:calculate_joker({destroying_card = scoring_hand[i], full_hand = G.play.cards})
-    if destroyed then break end
-end
-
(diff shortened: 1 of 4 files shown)


Catzzadilla:
man I sure do hope this gets in

PR #749, opened 2025-06-04
Fix `calculated` return in `context.modify_hand` not being returned

nh6574:
This allows for The Flint-like effects to get the proper animation.

## Additional Info:


nh6574:
Hmm, sorry after testing I realized that this will always juice up the blind chip rather than the source. So I feel like this need a more robust solution (or just doing it manually I guess)

PR #751, opened 2025-06-06
Fix DeckSkin Card UI (Crediting)

HuyTheKiller:
This addition would break modded languages on game launch with said language selected because there was nothing to stop the recursion, hence causing the "loop in gettable" error.

## Additional Info:

### DIFF
--- a/lovely/deck_skins.toml
+++ b/lovely/deck_skins.toml
@@ -240,3 +240,28 @@ else
 end
 '''
 match_indent = true
+
+#=======================#
+# DeckSkin Crediting UI #
+#=======================#
+[[patches]]
+[patches.pattern]
+target = "functions/UI_definitions.lua"
+pattern = '''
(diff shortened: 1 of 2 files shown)


Eremel:
I don't know enough about DeckSkins to know how this would affect it, @Kekulism can you take a look?

HuyTheKiller:
#659 has a DeckSkin feature to include artist credits as hover text in Customize Deck option, but it requires initializing `G.collab_credits` early. The commit uses a local recursive function to scan for the "Colalbs" label in UI tree and call the init function there, all inside a hook to `Game:prep_stage()`. It sure works for vanilla languages, but for some reason modded ones cannot stop the recursion at all. To circumvent that, I use lovely patches to inject an init variable so that when the hook pulls the Credits overlay, it instantly opens the Collabs tab.

PR #761, opened 2025-06-14
Fixed post_trigger effects not working on retrigger jokers

Git-i:
## Additional Info:

context.post_trigger effects don't trigger during calculate repetition even though there is trigger_effects call there, I've made some changes to fix it

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -1551,10 +1551,25 @@ SMODS.calculate_repetitions = function(card, context, reps)
         for _, _card in ipairs(area.cards) do
             --calculate the joker effects
             local eval, post = eval_card(_card, context)
-            if next(post) then SMODS.trigger_effects({post}, card) end
             for key, value in pairs(eval) do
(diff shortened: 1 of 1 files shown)


ItsGraphax:
Please look at this :)

Eremel (commented):
This doesn't quite have the correct functionality unfortunately, though it's certainly a good start.

The issue arises when a joker is set to do something outside of the return table in the post_trigger context, as the calculate function is only called one time. Take Hanging Chad, for example, where 2 repetitions are added. If I have a post_trigger that increases an internal value on my joker by 1, it'll only increase once but evaluate any return effects twice.

I think the best way to get around this would be to move where the post triggers are calculated to be where the repetition effects play, but I'm not 100% sure right now.

Git-i:
Oh I see, I'll edit the pr and see if I can get it working

Git-i:
I have pushed the fix for the issue, not sure if it automatically updates the pr, but it seems that way

PR #776, opened 2025-06-23
Chip-Mult Operator API

balt-dev:
## Additional Info:


balt-dev:
Meant to mark as a draft PR. :P

balt-dev:
Wiki PR made!

balt-dev:
Obviously, this is inspired by Jen's Almanac doing this first. I didn't take any code from it, however, so we're good on the code licensing front.

lord-ruby:
there should be a way to change what the operator displays as without defining node_func, with the option to define node_func still if you need something more complicated. having to do node func for every single operator seems like a bit much

gimmickCellar:
i think this should be added

balt-dev:
> there should be a way to change what the operator displays as without defining node_func, with the option to define node_func still if you need something more complicated. having to do node func for every single operator seems like a bit much

I could add an alternative way of specifying it that makes one for you, like for example
```
text = "X",
colour = G.C.UI_MULT
```
would autogenerate a `node_func` for that

balt-dev:
> there should be a way to change what the operator displays as without defining node_func, with the option to define node_func still if you need something more complicated. having to do node func for every single operator seems like a bit much

@lord-ruby You can now do this:
```lua
SMODS.Operator {
    key = "base",
    func = function(chips, mult) return chips * mult end,
    node_func = SMODS.operator_func("X", G.C.UI_MULT)
}
```

lord-ruby:
Im looking back onto this and one other thing that this could do with in passing self (aka the current operator) into both the node func and the regular func so that one operator could vary and so that a generic operators that change based on unpredictable things can be created more easily, base could be renamed to multiplication to make it more clear even if its pretty clear already

lord-ruby:
and some way to pass arguments into set_operator thats standardized get passed onto the operator itself for this same purpose would be pretty good too although idk how this will actually manifest

balt-dev:
Why did it close.

balt-dev:
Why are my changes no longer updating here. What is happening.

balt-dev:
Okay, I guess I somehow broke the PR entirely. Let me make a new one.

balt-dev:
See PR #809.

PR #779, opened 2025-06-25
Make blind's tag object accessible outside of skipping

the-Astra:
Allows access to a blind's associated tag (if any) via setting_blind context 

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -854,7 +854,7 @@ end
 position = 'at'
 match_indent = true
 payload = '''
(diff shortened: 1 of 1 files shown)


the-Astra:
Made a wiki PR here https://github.com/Steamodded/Wiki/pull/41

Eremel (commented):
Happy to merge the global variable once you mark as ready, but I don't feel like it's a relevant value to add to the setting_blind context.

the-Astra:
I do still think in one way or another, having access to a blind’s tag outside of skipping would be useful. Would it be better if I changed this to not include the context and just be the global game variable?

PR #781, opened 2025-06-27
Added optional scale context

VivianGiacobbi:
# Features
- Added an optional feature to enable contexts to be sent when values on cards scale
  - Scaling is determined through declared keys in `SMODS.valid_scaling_keys` for the `card.ability` table, or any value in `card.ability.extra` or its deeper tables not denied in `SMODS.denied_extra_keys`. When a valid keyed property is changed,`context.card_scale` is sent
- Receivers of the context can respond with the following keys:
  - `scale_message` responds to an expected scale message
  - `prevent_scale` prevents the incoming scale from occurring
  - `overwrite` overwrites the property instead of the incoming assignment
  - `add` modifies the incoming scale additionally
  - `mod` modifies the incoming scale multiplicatively
- Status text messages on scaling card are automatically modified, overwritten, or canceled on a card-by-card basis based on the results of the `card_scale` context.
  - If `prevent_scale` is returned for a given card, the message and any associated feedback is canceled
  - if `overwrite` is returned with a value, that value is used for the message
  - if `mod` or `add` are returned, messages are automatically altered to account for the value changes

# Optional
- This feature is optional due to increased performance cost for otherwise simple property accesses, as well as the feature demanding some more specific coding standards to make the best use of it
  - If you scale a relevant value, always message it immediately afterwards. Vanilla doesn't do this with Yorick's discards, which is why its key is left out

## Additional Info:


justamirror:
+1 i've had to do something like this for my own mod and having it be a part of SMODS would be great

PR #783, opened 2025-06-28
fix SMODS.destroy_cards(CardArea.cards) crashing

HuyTheKiller:
Destruction calls are put inside an event to ensure cards[i] won't go nil early, except for when events are messed up (e.g. by Nopeus' fast forward). Reverting the direction solves the issue.

## Additional Info:


HuyTheKiller:
#784 already has a better way to handle destruction calls (and also better support for eternal cards).

nh6574:
Does my PR solve the problem? I intentionally didn't make the same change to not conflict with this one.

HuyTheKiller:
Not really. Nopeus' Fast-forward just won't crash the game if destruction calls are immediate. You can still revert the loop direction in your PR.

PR #784, opened 2025-06-28
Changes to `SMODS.destroy_cards`

nh6574:
Adds the following parameters to `SMODS.destroy_cards`

1. `bypass_eternal` along with an Eternal sticker check. This could be extended in the future to support Eternal-like effects.
2. `immediate` to remove the event in the destruction call. Personally, I have some effects that I need to be run after the event inside start_dissolve/shatter but the event that wraps them was messing up with the timings.

## Additional Info:


nh6574:
Would be useful with this PR if both get merged https://github.com/Steamodded/smods/pull/794.

Issue #756, opened 2025-06-09
Using smart_level_up_hand() on a "???" hand in context.before does not properly update hand text


JustOsquo:
To reproduce:
Call `smart_level_up_hand()` in `context.before` while the hand text displays "???" (due to cards being face-down)

Expected result:
The text should be set to what the played hand actually is, as it normally would.

Example:
Playing a straight where at least one played card is face-down, causing the hand text to display as "???", and calling `smart_level_up_hand()` will prevent the hand text from properly updating to "Straight Lvl. x" once the cards are all face-up.

Likely cause:
`smart_level_up_hand()` does not take into account that the values in the hand text may not always be according to the actual hand played, even during a blind.

PR #800, opened 2025-07-05
Fix smart_level_up_hand() not properly updating hand text for hands with face-down cards

JustOsquo:
Calling `smart_level_up_hand()` while the hand display text is set to '????' due to cards being played while face-down would cause the function to incorrectly re-set the hand display text values after leveling up the given poker hand. This PR fixes the issue.
Relevant issue: #756 
## Additional Info:


JustOsquo:
On closer inspection, (and after noticing yet another issue with the old method, this time in regards to royal flushes) it is my belief that this alternative way of detecting the played poker hand is just way less broken overall, so I'll add a commit that just replaces it outright.

PR #804, opened 2025-07-06
Add context.money_altered (and G.STATES.SMODS_REDEEM_VOUCHER)

JustOsquo:
- Calculates `context.money_altered` when money is earned or lost by hooking `ease_dollars()`.
- I've also included `G.STATES.SMODS_REDEEM_VOUCHER = 998` for when a voucher is redeemed (since previously it would be considered `G.STATES.PLAY_TAROT`, which made it very difficult to separate the two)

`context.money_altered` also includes some built-in checks for roughly determining the source of the change in money,
(ex: `context.from_shop` checks for shop transactions)
though these are not strictly necessary.
```
if context.money_altered then
{
	amount = mod,
	from_shop = (G.STATE == G.STATES.SHOP or G.STATE == G.STATES.SMODS_BOOSTER_OPENED or G.STATE == G.STATES.SMODS_REDEEM_VOUCHER) or nil,
	from_consumeable = (G.STATE == G.STATES.PLAY_TAROT) or nil,
	from_scoring = (G.STATE == G.STATES.HAND_PLAYED) or nil,
	money_altered = true,
}
```

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -1808,3 +1808,33 @@ not self.ability.eternal then
 payload = '''
 not SMODS.is_eternal(self, {from_sell = true}) then
 '''
+
+# context.money_altered
+[[patches]]
+[patches.pattern]
+target = 'functions/common_events.lua'
+match_indent = true
+position = 'after'
+pattern = '''
+if instant then
+    _mod(mod)
+else
+    G.E_MANAGER:add_event(Event({
+    trigger = 'immediate',
+    func = function()
+        _mod(mod)
+        return true
+    end
+    }))
+end
+'''
+payload = '''
+if not instant then delay(0.4) end
+SMODS.calculate_context({
+    money_altered = true,
+    amount = mod,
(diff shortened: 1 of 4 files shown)


burnerboblmao222222:
This is super useful for a mod I'm making, thanks! I have spotted a few issues:
1. If clicking "buy and use" from the shop (eg. planet cards, some tarot / spectral cards), the purchase action of the card is not considered `from_shop` but instead `from_consumeable`. 
2. When in the booster pack choice interface, using cards like Temperance, The Hermit or Wraith will not be considered `from_consumeable`, but instead `from_shop`. (This also includes opening a booster pack from a tag and if the card comes from the pack or your consumables slot.)
3. Selling a card from the shop or booster open interface will be considered `from_shop`, but selling a card while in the blind will have all 3 `from` values be nil. Not sure if we'd need additional context parameters to capture this case.

(Note about 1: when clicking Buy and Use from the shop on Temperance, The Hermit, or Wraith, no. 1 still applies, but the additional action the card makes will fire correctly and be `from_consumeable`, which is correct.)

PR #817, opened 2025-07-15
nil check on CardArea.cards

HuyTheKiller:
Info about this issue can be found in [this Steamodded post](https://discord.com/channels/1116389027176787968/1209564621644505158/1394572105626157110) in Balatro server.
## Additional Info:


HuyTheKiller:
This is actually caused by JokerDisplay, which is already fixed in its latest commit.

PR #828, opened 2025-07-21
`SMODS.is_in_pool()`

nh6574:
Adds a helper function other mods can hook to check whether an object is in the pool.

## Additional Info:


nh6574:
Would it be nice to also make args an empty table when it's nil?

PR #838, opened 2025-07-27
(Moved to #1351 !) Implemented Quantum Ranks optional feature 

AllUniversal:
+Title

I gave implementing quantum ranks a try and am quite happy with the result.

Currently this supports all hand detection stuff (`X same` and `Straight`), I don't know how inefficient it gets but for vanilla hands it should be fine. (I tested 5 cards + 13 ranks per card and didn't notice any lag, and 7 vanilla cards with no lag either. Will do further testing if needed.)
`Shortcut` and `Four Fingers` are also supported.
Edge cases may still come up but some are already accounted for. (Like dealing with `rank.straight_edge = true` properly or forking paths in `.next`/`.prev` of a rank)

If you use the added `Card:is_rank()` function, this also supports any Jokers/Enhancements/Etcs. changing a card's rank, because I added a new `context` for getting a card's ranks (`context.get_ranks`). 
-> Importantly, I have not (and probably will not) implement cards holding more than one rank _independently of any effects_. As in: If you want specific cards to have more than one rank, with this implementation you'd have to make a custom `Enhancement` or similar, which returns the intended ranks during `context.get_ranks` evaluation for the card it's applied to.

I tried commenting my code but further documentation will follow (like a PR to the Wiki repo).

Patching the vanilla Jokers could use some attention by a regexpert (<- not me!).

Any feedback is welcome! I fully expect my straight algorithm (especially) to be replaced by something better eventually, but for now I just wanted to get something working. :)

## Additional Info:


AllUniversal:
Just pushed an optimization which in theory should drastically reduce straight calculation times for hands with many forking / branching straight paths. Still gotta test it though.

AllUniversal:
I realized that get_X_same will actually need another look, as right now cards can be counted multiple times, which at the very least is inconsistent with how they work in straights. I'll try fixing it.

AllUniversal:
Alright, the above issue has been fixed and `get_X_same()` now works just like straight calculation. 
This required changing how `Full House` and `Flush House` check their respective `PokerHandParts` and also how those call `get_X_same()`, specifically, only `_5` and `_all_pairs` now call the function with `or_more=True`. (`_2`, `_3` and `_4` no longer do)

AllUniversal:
Alright, I tested 5 cards with each card counting as its own rank and its .next rank (picture below). No lag yet on my end (really shouldn't be lagging yet).

image
image

Straight Flush with 7 cards: 

image

Flush House with 8 cards: (2 Jacks, 10->Jack, 9, 8-> 9 => J J J 9 9

image

Performance wise it seems fine, but I don't know why the second to last hand (straight flush) doesn't evaluate as a Flush House.. 
Will look into it.

AllUniversal:
`Full House` calculation might need its own function. I'll give it some more thought first though.

nh6574:
Nice work on this!! I'm excited to see this merged

Though I would recommend turning it into a draft PR if you're still working on it

AllUniversal:
image

Fixed the issue! (Hopefully) (Surely)

AllUniversal:
Imma mark as ready now because I can't think of any more additions I'd want to add.
Actually let me make a list of remaining issues (that I'm aware of):

- [ ] `Royal Flush` detection relies on `card.base.value`, but changing that would require running a recursive check or something, which to me does not seem worth it.
- [x] Performance with really large hands and or a lot of custom ranks hasn't been tested. (EDIT: 2025.8.13, I have now tested performance up to 20-35 cards in different conditions. See below comment)
- [x] In fact, custom ranks haven't been tested (but _in theory_ shouldn't cause any problems as long as their `.next` is correctly defined) (EDIT: 2025.8.10, my recent changes have refined `Rank.straight_edge` to now (in theory) allow Paperback-Apostle-like rank implementation with straights. Still have to test them though.) (EDIT: 2025.8.21, I have now reimplemented Paperback's Apostle, straight calc seems to be working!)
- [ ] Documentation and perhaps more code commenting (Once this gets approved/merged, I'll make a Wiki PR)
- [x] I forgor about Stone Cards woopsie
- [x] Even Steven, Odd Todd and Cloud Nine still aren't using Card:is_rank() (EDIT: 2025.8.24, Even Steven and Odd Todd now use the new `Card:is_parity()` function. With this Imma mark this as done, because I ain't implementing Cloud Nine)
- [x] Card:is_face() compatibility
- [x] no_mod context argument for better control

In my opinion, none of these (except maybe performance) are critical. The vanilla Jokers would be nice, but Cloud Nine really reasonably can't use Card:is_rank(), because the ranks of a card are entirely context dependent, so leaving it as get_id() instead makes sense to me.

AllUniversal:
Realized I need to implement some more stuff, updated above message.

Edit: Alright done. Will mark as ready again. :)

Edit 2: Aaand there's an issue with `Full Hand` calculation. I'mma go cry and then fix it!

Edit 3: Nvm my 'all' branch didn't contain all changes for some reason lmao

AllUniversal:
I've marked this as Draft again because I wanna wait to see whether #877 gets merged, and I'll clean up some stuff too probably.
(+ I should probably test custom ranks)

AllUniversal:
I tested a simple implementation of Ortalab's "Scenic Route" Joker (which allows Straights to contain pairs),
should be as simple as this;

image

image

AllUniversal:
Above commit added a new optional flag: `SMODS.optional_features.quantum_straight_min_return` 
This flag enables an optimization which greatly decreases straight calc time, **but does not ensure that the longest straight is returned**.
However; As long as there is only one straight in the played hand, and as long as there are no one-way straight_edges branches (like Paperback's Apostle causes), it _**will**_ return the straight **fully extended**.
It will also never fail to find a straight, if there is one.

Using this is only necessary with ~14+ card hands, depending on other factors (Shortcut, Ranks-per-card, Ranks-per-.next/.prev).

As for making it a new `SMODS.optional_features` flag, I'm unaware of a better place to put this, though I'd personally prefer not to add unnecessary optional features.

I'll post a short list of performance tests soon.

AllUniversal:
I have now tested performance with 20 card hands;
_Note: My PC, though not extremely high end, isn't a Laptop either, so results may depend on your device_

1. Vanilla, 24 cards, `quantum_straight_min_length = false`: 
    No lag
2. Vanilla, ~21 cards, Shortcut, `quantum_straight_min_length = false`: 
    No lag, then stuttering/crash when selecting more cards
3. 13-rank cards, ~15 cards, `quantum_straight_min_length = false`: 
    No lag up to 14 cards, then suddenly crashes when selecting the 15nth (?!?)
    _(Note; As a Wild Rank implementation, this is inefficient. Instead you should implement a Wild Rank tally into straight eval,
    which is then used as an "allowed skips counter", (and Wild Rank cards are excluded from calculation))_
4. Vanilla, 35+ cards, `quantum_straight_min_length = true`:
    No lag.
5. Vanilla, 35+ cards, Shortcut, `quantum_straight_min_length = true`:
    No lag.
6. 13-rank cards, 35+ cards, Shortcut, `quantum_straight_min_length = true`:
    FPS starting to drop a little, barely any lag though.

I'm now suspicious of the code I have written. But I don't think there's any mistakes, the new optimization simply stops once it hits the end of the first valid straight it finds. This means in large hands like these, if the final evaluated poker-hand is a straight, it might pick a 5 card one or a 20 card one, or other, depending on which was found first.

AllUniversal:
Having updated Paperback's Apostle, seems to be working as expected;
image
image
And "Rapture" Straight Flush;
image

This is the new Apostle code;
image
(+ The extra wrap around test / `get_straight()` hook can be removed)
(EDIT: Above code is missing `next = {"2", "paperback_Apostle"}` in the `Ace:take_ownership` call (I think..? Maybe?))
(EDIT 2: And why did I write `prev = {"**2**"..` ?! Silly me)

AllUniversal:
With the above custom rank test, I'm going to re-ready this now.

AllUniversal:
Alrighty NOW I'm ready for review (fr fr, v2-final-final-last-1.0).

Above commits fixed some minor edge cases, a fix for a problem with Shortcut I thought I had accounted for but actually didn't (woopsie), and **added Wild Rank API stuff**.

Such as; `SMODS.has_any_rank()` hookable to define your own behavior, supports Enhancements, Editions, Seals and Stickers by default (set `thing.any_rank = true`).

Using this should be much more efficient than just returning all ranks in `context.get_ranks`.

AllUniversal:
I've once again realized that the current `straight_edges` implementation causes some unlikely edge cases. I don't know if I'll fix them though, because they only occur with custom ranks, in very specific (and as mentioned, unlikely) scenarios. 
In case this gets merged, if such issues are encountered frequently I'll look into it, but until then I can't get myself to redo straight calc again. :')

EDIT: Specifically;
If a custom rank supersedes another custom rank which then connects to the Ace, and each defines it's `.next` in its straight_edges, a (somewhat) faulty optimization may `return` early, depending on which cards are checked first during eval.

To fix this, I could change `straight_edges` again to work as follows:
`Ace => {King = true, 2 = true}`  ---> `Ace => {King = {2 = true}, 2 = {King = true}}`
This means that _coming from a 2 into an Ace_ disallows passing _specifically_ to a _King_ and vice versa.
In theory this way of doing straight_edges is more versatile, but it also allows defining behaviour which isn't supported;
Using this structure, define a custom rank 'F' that passes through the Ace, to another custom rank 'G'. Define it in the Ace's straight_edges, so that a straight may not pass from 'F' to 'King', or from 'G' to '2'. This means there may be a straight that contains an Ace in its middle using these and further custom ranks. 
The current implementation will split a straight starting from a rank that has _any_ straight_edges defined. So if these custom ranks themself _also defined straight\_edges_, the valid straight through the Ace wouldn't be found.

Looking at this, I may still need to update the straight_edges implementation to avoid the existing issue with consecutive custom ranks.

src/overrides.lua
```diff
@@ -593,74 +593,457 @@ G.FUNCS.change_stake = function(args)
...
+			if or_more and (tally >= num) or (tally == num) then
+				table.insert(ret, rank_cards[rank])
+			end
+		end
+		return ret
+	end
 end
 
+
+function get_quantum_full_house(_3, _2)
```
Aurelius7309, line 946: This and `get_quantum_two_pair` are too specific and require a lot of extra effort for hand types that combine multiple sets of some amount of cards each (think a fuller house 4+2, two triple, three pair, etc.). There needs to be a utility that can do this in the general case. It'd accept a combination in form of a table like `{length:amount}` and the corresponding hand parts indexed by their length. I can give implementing this a shot myself if you don't want to.

src/utils.lua
```diff
@@ -2684,6 +2709,116 @@ function SMODS.is_eternal(card, trigger)
...
+
+    local ret = {}
+    for _, v in pairs(eval.ranks) do
+        if v then ret[#ret+1] = v end
+    end
+
+    return ret
+end
+
+function Card:is_parity(parity, exclude_faces)
```
Aurelius7309, line 2813: Again, face cards shouldn't be special in regards to parity - they inherently don't have one. So there should be no need to explicitly exclude them here

src/overrides.lua
```diff
@@ -593,74 +593,457 @@ G.FUNCS.change_stake = function(args)
...
+							for _, n_rank in pairs(rank_nexts) do
+								local rec_ret_straight = recursive_get_straight(n_rank, next_straight, max_skips, do_wrap, direction, used_c_reps, rank)
+								if #rec_ret_straight > #ret_straight then
+									ret_straight = rec_ret_straight
+								end
+								if #ret_straight >= #hand then
+									return ret_straight
+								end
+							end
+							if SMODS.optional_features.quantum_straight_min_return and #ret_straight > min_length then return ret_straight end -- Greatly decreases calculation time but doesn't ensure that the longest straight is found.
```
Aurelius7309, line 819: This optional feature is terrible in terms of API design. If this optimization is at all to be left up to individual mods, then what a mod may require is for it to not be used because it's important for its mechanics that the longest available straight is always detected correctly. This means the optimization should be opt-in instead of opt-out - otherwise such a mod will have no way of ensuring longest straights being found. Even then it might be more of a decision for the player to make instead of for specific mods to mandate?

lsp_def/classes/rank.lua
```diff
@@ -17,7 +17,8 @@
 ---@field next? Ranks|string[] List of keys to other ranks that come after this card. 
 ---@field prev? Ranks|string[] List of keys to other ranks that come before this card. Used when evaluating straights. 
 ---@field strength_effect? table|{fixed?: number, random?: boolean, ignore?: boolean} Determines how cards with this rank behave when Strength is used. 
----@field straight_edge? boolean Sets if this rank behaves like an Ace for straights. 
+---@field straight_edges? table<Ranks|string, boolean> Sets the rank keys from which going into this rank should terminate a straight. (Vanilla only set for Ace => {'King' = true, '2' = true})
+---@field parity? number The parity of the rank, 0 for Even, 1 for odd, anything else for neither
```
Aurelius7309, line 21: "Anything else" is setting a bad convention, ranks with no parity should leave this blank

Aurelius7309 (changes requested):
Had a first closer look at this, here's what I've noticed without digging too deep into the technical side of it

src/overrides.lua
AllUniversal: Leaving it up to the player may make sense yeah. I don't like having it be a whole other `optional_features` flag anyway, that said it _is_ opt-in, right? You have to set it to true like the other optional features.
It is a little ugly either way. Problem is that I personally can't implement this in a way that allows 30+ card hands without using this optimization. If someone else wishes to take another look, be my guest :)
(However, for _most_ scenarios, the longest straight should be found either way) (I think. Has to be tested before I'm 100% sure)

Aurelius7309: Whoops! I'm getting tired I guess. What I meant to say is that the optimization should be opt-out instead of being opt-in if it's left up to mods, so that any mod can decide it wants the optimization to be off, as the mods that want to leave it on likely won't have any issues other than performance with it off

Aurelius7309: No performance related ones, anyway. If there's other performance settings that could be added, this could go along with them in a new tab/page. Seems fine for now as it's the easiest thing to change

src/overrides.lua
```diff
@@ -593,74 +593,457 @@ G.FUNCS.change_stake = function(args)
...
+					for _, p_rank in ipairs(pcard_ranks) do
+						ranks[p_rank.key][card_rep] = true 					-- Add card_rep to rank
+					end
+				end
+			end
+		end
+
+		-- TODO: Optimization that may or may not be worth it: 
+		-- Remove all card_reps that share a rank with a 1-rank card_rep at that rank. Repeat until none are removed.
+		-- Also, collapse any N card_reps that share N ranks into one of those ranks each. (Hard to implement / too inefficient maybe)
```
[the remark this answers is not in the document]
AllUniversal: Yeah, doing it for all card_reps / all ranks probably would be too much. I kinda hate this optimization :) Because straight calculation isn't perfect anyway probably. So replacing it if it performs badly seems better anyway. That said, I first gotta know how badly it performs.

src/overrides.lua
```diff
@@ -593,74 +593,486 @@ G.FUNCS.change_stake = function(args)
...
+	for i, v in ipairs(parts) do
+		parts_list[i] = v
+	end
+
+	local valid_pairs = {}
+	local used_cards = {}
+	local used_pairs = {}
+	if base then
+		for _, pairs in ipairs(base) do
+			for pair, over_by in pairs(pairs) do
```
Aurelius7309, line 962: This is cursed, please don't call a local variable pairs

cg-223: pairs(pairs)

AllUniversal: Woopsie

Aurelius7309:
Could you provide some insight on how the get_ranks context is meant to be used to add ranks to cards? Simply returning the ranks you want added clearly strips the cards of any other rank(s) it previously had, but returning the ranks passed from context as well ends up not being deduplicated correctly:
grafik
grafik
This causes each 5 to be counted as its rank twice, which is apparently enough to make it a 4oak? Also, does it handle additional ranks coming in from multiple calculation targets - I can't seem to find any logic that suggests it does?

AllUniversal:
@Aurelius7309 Of course! 
The way I implemented it works like you mentioned; the returned `ranks` flag overrides the `context.ranks`, to add ranks you should be able to do what you did above or similar, so I gotta see why that isn't working...
Having multiple calculation targets add ranks should work fine as long as each is implemented correctly, right? `context.ranks` gets updated with the returned flags between calculations. (In `utils.lua:~1755`).

Importantly; The way I did this, Quantum Rank effects are inherently order dependent, so imagine this;
Joker A:    7s also count as 2s
Joker B:    2s also count as 4s 
If Joker A is eval'd before Joker B, played 7s have the ranks 7, 2 and 4.
If Joker B is eval'd before Joker A, played 7s have the ranks 7 and 2.
This is true depending on how Joker B checks whether the card is a 2;
True, if Joker B uses `context.ranks` as reference
False, if Joker B uses `Card:get_id()`
In the latter case, Joker B will only ever add `4` as a Rank if the played card is a vanilla `2`.

I can't think of a way of doing this that also allows removing/setting ranks and isn't ordered, but tell me if you have any suggestions.

I'll have to check why `unpack(context.ranks)` isn't working as I expect here.

Also, I have realized that there's a problem with infinite context loops (I believe you mentioned that on Discord).
When a Quantum Rank Joker calls `SMODS.has_enhancement()` while there's a Quantum Enhancement Joker that calls `Card:is_rank()`, the contexts call each other. (The Jokers/Calc Targets technically don't even have to be separate).
I have thought of a way to (hopefully) fix this and I'll make a separate PR soon™️.

Oh and also^2, 
'2', 2 or SMODS.Ranks['2'] are accepted as `ranks` in `get_ranks`'s flags, so you don't _have_ to get the `SMODS.Rank` for every rank you want to add.

Aurelius7309:
I'd like to add that `context.ranks` being a list of rank objects is a bit impractical to work with (e.g. you have to iterate the entire list to check if something is to be considered a certain rank like "all 7s also count as 2s") and also inconsistent with the way quantum enhancements work. Working with tables indexed by rank keys instead seems more natural.
Oh and I tried just passing the keys to the returned ranks, but ended up crashing because the keys weren't converted to rank objects - maybe the conversion is going wrong somewhere?

AllUniversal:
You right you right, I'll change it to be a map instead of a list.
And I'll see about the conversion breaking :) Very strange.

AllUniversal:
Alright I changed it. Will do some more testing to see if stuff breaks. 
Also consider taking a look at this pretty please :) #875

Aurelius7309:
Thanks, I'll have another go at testing as well. I'd be in favor of changing Card:get_ranks() returns as well for consistency, are there any reasons against this?
#875: will do :)

AllUniversal:
I've decided to redo the `get_ranks` context to use two accepted return values instead of just `ranks = {...}`.
It now accepts `ranks_fixed` and `ranks_changed`. Examples;
Returning `ranks_fixed = {['5'] = true}` would make all cards count as only 5s.
Returning `ranks_changed = {['5'] = true}` would make all cards additionally count as 5s.
Returning `ranks_changed = {['5'] = false}` would make all cards NOT count as 5s. (Assuming later calculation targets don't override this)

Edit: In reply to @Aurelius7309 
Yes, I am lazy and don't want to adapt the 7 occasions where `get_ranks()` is called :) 
Just kidding I'll change it too.

Aurelius7309:
> Yes, I am lazy and don't want to adapt the 7 occasions where `get_ranks()` is called :)
> Just kidding I'll change it too. 

Haha I figured as much that it'd be a bit of a hassle to change, but rather change it now than be stuck with something bad later :p

> I can't think of a way of doing this that also allows removing/setting ranks and isn't ordered, but tell me if you have any suggestions. 

Forgot to comment on this earlier, depending on evaluation order is totally fine as all effects work this way

AllUniversal:
#945 This PR I have opened just now is also important for Quantum Ranks, it addresses a big issue with calling `SMODS.has_enhancement()` or `Card:is_rank()` during `check_enhancement` or `get_ranks` contexts.

AllUniversal:
I've thought a little about my current `straight_edges` implementation and realized that the proposed `VirtualRank`s could work pretty well. As in, create two such `VirtualRanks` for the Ace, `Ace_high` and `Ace_low`, have them update/override the Ace's nexts and the King and 2's respective .next/.prev too (only for straights though, so probably store these virtual `.next/.prev`s in the `SMODS.Rank`s). 
Among other changes this would pretty much require `.next` to be turned into a map, which would in turn require `strength_effect` to be reworked/fixed.
For now, I'm too tired to try again. Maybe later :)

AllUniversal:
The recent commit contains some breaking changes; I've replaced `straight_edges` entirely, in favour of a new class, `SMODS.VirtualRank`. It allows for creating rank "redirects" specifically for the purpose of straight calculation. 
By default this includes the `VirtualRanks` `Ace_high` and `Ace_low`, which appropriately redirect the `King` and `2`'s `.next`/`.prev`s and replace the `Ace` for straight calc.

To make this easier **I've changed `.next` and `.prev` to be maps instead of lists**, this means that;
A. their definitions have to be adjusted for custom ranks
and B. it is now recommended that each Rank defines its `strength_effect` explicitly, with the new `fixed_next` and `fixed_prev` fields set to the appropriate rank keys.

Though `VirtualRanks` add quite a bit of complexity, they also avoid many of the problems the simpler `straight_edges` implementation had. (I think (: )

(EDIT: Oh and also rn the vanilla `get_straight()` is commented out, I gotta see if I can just delete it entirely because I haven't updated it yet)

AllUniversal:
Alrighty for what may hopefully be the last time, I've reworked `get_straight()` once again.
Its recursive function now correctly backtracks between `direction = next` and `direction = prev`, this also comes with some massive syntax simplifications. 
Though this may have a slight negative impact on performance, it now should guarantee that straights are found correctly, even in specific circumstances which previously would have caused problems (due to an important optimization which caused an edgecase).
Those circumstances being; If a quantum-rank-card's ranks appeared both before and after (next and prev) the starting rank, and a fork in the available ranks exists out of view of the current straight, in large-number-of-cards-hands, this may have caused no straight to be found. It no longer should.

EDIT: After some more thinking,
this won't be the last changes to `get_straight()`.
:')

AllUniversal:
Here's my updated ToDo list (4$ per todo completed):

- [ ] Performance checks / Replace straight calc with bipartite matching algorithm (I don't know what that means)
- [ ] Custom Rank checks (should work flawlessly™️ now in theory)
- [ ] Documentation / Complete changelog
- [x] Stretch goal: Add `SMODS.contains_cards()` function for generalized 'does this list of cards contain a card with rank X and enhancement B, and a card with rank Y and seal G, etc..' -> This would make custom poker hand types much much easier if they rely on specific ranks/enhancements/etc. (EDIT 2, 2026.4.1 This has now been fully implemented, no performance tests done, but it shouldn't be too bad.)
- [x] Stretch goal+: Royal Flush compat with quantum ranks (probably won't happen) (EDIT: As suggested below, I have implemented a simple `is_royal` field for Ranks. This will do for now, though it does lead to some quirky labeling of Straight Flushes as Royal Flushes, but better than nothin'.)
- [x] Stretch goal++: Cloud Nine (almost definitely won't happen, at least not without reworking Cloud Nine's update timing)

Aurelius7309:
Re: Stretch goal+: this requires some further thought on what a generalized royal hand is. Maybe it makes sense to introduce another property that sets whether a rank is to be considered "royal". I think it then makes the most sense to consider the hand royal if all scored cards have at least one royal rank. This shouldn't be too hard to then extract into a neat helper function to be used by other royal variations of hands. I did also just notice that we have a minor inconsistency in Royal Flushes compared to vanilla, namely that of rankless cards being handled incorrectly: stones that have an underlying royal rank don't invalidate a Royal Flush in smods whereas they do in vanilla.

Re: Stretch goal++: This could either benefit from the vanilla update behaviors to be made lazy, or I'm fine with explicitly converting Cloud Nine to only check in loc_vars and the relevant calc contexts. Both options are fragile wrt game updates, but that's acceptable for the sake of functionality.

Aurelius7309:
Re: Cloud Nine: Card:update() does not return automatically when a method is present on the center. I'm not sure why, but if there isn't a mod that relies on adding update functionality to vanilla jokers with the original still being run (which I doubt, and it would be bad practice anyway), that behaviour can easily be changed. Otherwise I suggest a patch to disable the relevant part of the vanilla function. As is, your fix should work but won't remove the original overhead that is no longer needed.

soulware1:
i have no clue how these work but, maybe we need some kind of function to get the lowest/highest possible ranking in a table of cards?

AllUniversal:
Been a while! :) Haven't done any work on this, _until today_! 
Just added a set of new `SMODS.match_cards()`-and-related functions; They allow for matching cards by certain conditions, like seals or enhancements + combinations of conditions. 
(This is the hook I used for testing): 
image
`SMODS.match_cards()` returns a map of the matchers (a matcher is a collection of conditions) indexing a map of cards each matcher matched with, and the inverse, so a map of cards indexing a map of each matcher a card matched to. This already allows testing whether each card matches a matcher and the reverse, but doesn't yet support de-duplicating such matches. (Like had to be done with quantum straights, basically the same thing here, probably requires implementing a general bipartite matching function or something.)

AllUniversal:
Just added one (probably last) big feature; `SMODS.get_hand_from_matching()`, a function which allows de-duplicated pokerhand detection, as in, it can do the following;
image
With this code it detects "Flushes" like this; 
image
But not this; 
image
EDIT: This might best be a separate PR actually. LMK if I should split it off.

src/utils.lua
```diff
@@ -315,35 +315,33 @@ function SMODS.modify_rank(card, amount, manual_sprites)
...
     if amount > 0 then
         for _ = 1, amount do
-            local behavior = rank_data.strength_effect or { fixed = 1, ignore = false, random = false }
+            local behavior = rank_data.strength_effect or { fixed_next = false, ignore = false, random = false }
             if behavior.ignore or not next(rank_data.next) then
                 break
             elseif behavior.random then
                 rank_key = pseudorandom_element(
-                    rank_data.next,
+                    SMODS.get_keys(rank_data.next),
```
Aurelius7309, line 323: This creates inconsistent RNG due to `pairs` having undefined order

src/utils.lua
```diff
@@ -1931,13 +1955,58 @@ function SMODS.update_context_flags(context, flags)
     if flags.denominator then context.denominator = flags.denominator end
     if flags.cards_to_draw then context.amount = flags.cards_to_draw end
     if flags.saved then context.game_over = false end
+    if flags.ranks_fixed or flags.ranks_changed then
```
Aurelius7309: This construct is awkward to use. Using keys over rank objects seems like the better choice, given it spares you from having to type out `SMODS.Ranks` all the time when you don't care about any properties of the rank object. It also seems more usable to have a single return key for changed ranks (this can just be `modify`) and a boolean flag like `fixed = true` if you want to override existing ranks

src/overrides.lua
```diff
@@ -619,74 +619,424 @@ G.FUNCS.change_stake = function(args)
...
+		for _, valid_pair in ipairs(valid_pairs) do
+			local valid_hand = get_quantum_pairing(parts_list, valid_pair)
+			if valid_hand then
+				return valid_hand
+			end
+		end
+		return nil
+	end
+
+	if next(valid_pairs) then
```
Aurelius7309, line 1024: This merge results in duplicate cards in the merged hand if cards have more than one rank that each belongs to a different pair. I'm unsure if this causes any issues, but deduplicating this is simple and the safer choice

src/utils.lua
```diff
@@ -2992,6 +3084,176 @@ function SMODS.is_eternal(card, trigger)
...
+            if all and not is_wild then return false
+            elseif all then is_wild = false end
+        end
+    end
+    return all
+end
+
+function Card:get_ranks(flags) -- Returns a map of "SMODS.Rank"s, sanitized to ONLY be "SMODS.Rank"s -> Rank keys or rank ids are converted to SMODS.Rank 
+    if not self.playing_card then return {} end
+    local default_ranks = (SMODS.has_no_rank(self) and {}) or {[SMODS.Ranks[self.base.value]] = true}
```
Aurelius7309: `no_rank` is weak, causing `SMODS.has_no_rank` to yield inconsistent results if any quantum rank effects adds a rank to an otherwise rankless card. Conversely, `SMODS.has_no_rank` doesn't reliably indicate if a rank is "effectively rankless". These interactions make the function quite unintuitive to use; perhaps it's better to differentiate between general no-rank checks and ones used in checking ranks?

src/utils.lua
```diff
@@ -2992,6 +3084,176 @@ function SMODS.is_eternal(card, trigger)
...
+    if type(cards) ~= "table" then return false end
+    for _, pcard in ipairs(cards) do
+        if not pcard:is_royal() then
+            return false
+        end
+    end
+    return #cards > 0
+end
+
+function SMODS.lowest_and_highest_rank(cards)
```
Aurelius7309, line 3218: What is this meant to be used for? Sort order is not generally indicative of one rank being higher/lower than another

Aurelius7309 (changes requested):

src/utils.lua
AllUniversal: Would sorting the keys alphabetically resolve this?

Aurelius7309: Assuming usage is restricted to string keys only, yes.

src/utils.lua
AllUniversal: I've changed `Card:get_ranks()` to now use a single `ranks` return field and the `fixed` flag as you suggested it.
Additionally it now only takes rank keys as return values.
Edit: Oh and I opted for `ranks` over `modify` for clarity + I'm currently looking at a general quantum API where this would probably be necessary.

src/utils.lua
AllUniversal: It's meant for getting the highest or lowest card, for effects like Raised Fist, but you're right to say that `sort_nominal` isn't really the best measure of "highest" or "lowest" with ranks. Any suggestions on what to replace it with? Might have to add a field to `SMODS.Rank`.
Edit: That said, I do believe `sort_nominal` is also what's used to sort the cards in hand, so there's some logic to using it here.

src/utils.lua
AllUniversal: Hmm, so should I move the current checks from `SMODS.has_no_rank()` over into `Card:get_ranks()` to decide the `default_ranks`, and then have `SMODS.has_no_rank()` use `Card:get_ranks()` to check whether a card has no ranks?

src/utils.lua
Aurelius7309: The issue with any approach is that a generalized set of ranks doesn't need to have a total ordering, so highest/lowest ranks as a concept may not make sense. In that case `sort_nominal` probably does make the most sense

src/utils.lua
Aurelius7309: I guess that makes sense. The alternative is an additional argument for `SMODS.has_no_ranks` — not sure right now what's cleaner

src/utils.lua
AllUniversal: I agree, guess I'll leave it as is then.

src/utils.lua
AllUniversal: With some recent thinking for a potential general quantum API, I realized there's a fundamental problem with having `SMODS.has_no_rank()` inside `Card:get_ranks()` to define the default values; 
`SMODS.has_no_rank()` currently supports quantum *enhancements*, in a general quantum API, those would *also* use a respective `SMODS.has_no_enhancement()` function in their getter function to define default values, but that causes a loop; 
1. The getter is called, 
2. has_no is called to define default values, 
3. has_no calls all/some getters for *other* quantum fields (e.g. seals, editions, enhancements, etc.), 
-> A loop is entered.

I'm currently thinking about how to approach a solution; It might be unavoidable to have `no_rank` and the like *not* support quantum enhancements/etc, thus e.g. a QE Stone card would *not* count as `no_rank`/`no_suit`.
Still looking into a better solution. (My brain has taken up a side-gig as an industrial oven.)

AllUniversal: I have now opened a new Draft PR #1351, it tackles this is issue by not letting QEs etc. affect `no_...`/`any_...` of a card. In theory there may still be a way to have it work, but I dunno. :)

AllUniversal:
I've now merged this PR into #1351! I will now be working on this and all the new stuff over there, so I'm going to close this PR.

PR #840, opened 2025-07-28
card and booster pack weight change functions

tomatoseandcrying:
Two new `Card:` utility functions that should provide more accessibility to editing object weights: 

`set_booster_weight()`
`set_card_rate()`

## Additional Info:

### DIFF
--- a/lsp_def/utils.lua
+++ b/lsp_def/utils.lua
@@ -2,73 +2,73 @@
 
 --- Util Classes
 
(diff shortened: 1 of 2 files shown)


english5040:
Before anything else, you need to un-autoformat some of your files

english5040:
These functions could be made more general: modifying the weight of new consumable types should be possible, and the system of G.GAME.X variables should probably be changed to use contexts like the new probability system

PR #850, opened 2025-08-01
Prevent calculate_context from being called outside of a run (Closes #849)

### LINKED ISSUE #849: Prevent calculate_context from being called outside of a run
What the title says. I know this is more of a problem on the modders' side but I'm seeing this problem pop up more frequently recently. It's usually caused by a mod calling their custom context or calling a function that calls a context (like `SMODS.is_eternal`).

nh6574:
See #849 for context.

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -1827,6 +1827,8 @@ end
 -- Used to calculate contexts across G.jokers, scoring_hand (if present), G.play and G.GAME.selected_back
 -- Hook this function to add different areas to MOST calculations
 function SMODS.calculate_context(context, return_table, no_resolve)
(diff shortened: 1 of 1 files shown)


AllUniversal:
Are the `get_probability_vars()` contexts a problem here (because they get called in the collection)? If the function just returns the params it received, should be fine I guess.

nh6574:
> Are the `get_probability_vars()` contexts a problem here (because they get called in the collection)? If the function just returns the params it received, should be fine I guess.

That one already has a check for not being in a run and does exactly that.

nh6574:
But maybe it should return an empty table in case something checks for flags and expects a table.

PR #852, opened 2025-08-03
Fix `strength_effect.random`

wingedcatgirl:
`strength_effect.random` crashes if any of the rank's `next` values don't have an `in_pool` method. Such as all vanilla ranks. Oopsie!

## Additional Info:


Eremel:
Oh wait this is already fixed with 760c2de

PR #618, opened 2025-04-07
Update zh_CN

snowylight:
## Additional Info:

- [x] I didn't modify api's or I've made a PR to the [wiki repo](https://github.com/Steamodded/wiki).
- [x] I didn't modify api's or I've updated lsp definitions.
- [x] I didn't make new lovely files or all new lovely files have appropriate priority.

PR #857, opened 2025-08-06
Update zh_CN.lua

HONE13:
Update translation


HONE13:
Update new text

english5040:
Can you take a look at the other open Chinese localization PR #618 to check if that one has any improvements? (Other than that we can probably merge straight away)

english5040:
this has been inactive for a bit, just going to merge it as it doesn't have code

PR #858, opened 2025-08-06
Update en-us.lua

HONE13:
Fix error text


nh6574:
This is not an error, the numerator now has different values due to the new probability functions.

PR #865, opened 2025-08-08
scale_card message related changes

lord-ruby:
Scale card now returns an optional message that is defined by returning a scaling_message in calc_scaling if the returned message is a non-string the original message will be prevented entirely

## Additional Info:


Eremel:
Covered in 85f953c

PR #871, opened 2025-08-09
Revert `ease_dollars` position

HuyTheKiller:
Previously `context.money_altered` was added as janky patches. It's changed to a hook now, but this part remains, causing `ease_dollars` actions to be delayed considerably.
## Additional Info:


Eremel:
This is necessary for correctly timing events that are added due to dollars being altered

HuyTheKiller:
You could've at least undone `ease_dollars` and accepted `SMODS.shatters`.

Eremel:
Don’t lump unrelated things together in the same PR then. This PR is about ease dollars so that’s what I’ll be looking at.

PR #876, opened 2025-08-10
calc_interest, and variables to better modify interest

bepisfever:
Adding 2 global variables that can be changed to modify interest more easily without modifying the behavior of vanilla interest-related variables.
- `G.GAME.extra_interest_cap`: Adds onto the current interest cap.
- `G.GAME.extra_per_interest`: Adds onto the amount of money per $1 interest.

`SMODS.get_interest()` is added, which returns the total interest cap, and total cash per interest.

And Jokers can now have `calc_interest(card)`:
- Returning `add_cap` will increase interest cap by that amount.
- Returning `add_per_interest` will increase cash per interest by that amount.

## Additional Info:


english5040:
Very reasonable, making interest easy to modify is a good idea. I remember the getter contexts thing; that seems applicable to this, you may need to update this PR.

About API structure: I am concerned about Card:calculate_interest(), because you're adding on another function onto every card. It feels a bit unnatural.

PR #907, opened 2025-08-22
Add apply, remove and set_ability functions to Seals

bepisfever:
Seals now have:
`apply(self, card)`: Runs when the Seal is applied on an obtained card, or when the card with the Seal is obtained.
`remove(self, card)`: Runs when the card with the seal is removed, or when the seal is removed.
`set_ability(self, card)`: Runs when the Seal is applied on a card.

I also added a hook that adds `added_to_deck` to all playing cards at the start of run.

Wiki PR update: https://github.com/Steamodded/Wiki/pull/53

## Additional Info:

### DIFF
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -2388,20 +2388,16 @@ end
 local card_remove_ref = Card.remove
 function Card:remove()
     local ret = card_remove_ref(self)
-    if self.seal and self.added_to_deck then
-        if G.P_SEALS[self.seal] and G.P_SEALS[self.seal].remove then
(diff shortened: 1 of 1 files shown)


DigitalDetective47:
It would be nice to be able to distinguish between calling cases. For example with removed: being able to differentiate the seal being replaced from the card with the seal being destroyed.

english5040:
as always, naming concerns: apply/remove/set_ability don't really match with what they actually do. I propose `add_to_deck`/`remove_from_deck` (like Jokers),  `set_seal`

PR #908, opened 2025-08-22
Better HEX function

bepisfever:
`HEX` can accept HEX strings with "#" at the start, like `#FFFFFF`, and can also accept RGB values in both `table` and `string` forms.
EX: `HEX{"255","255","255"}`, `HEX("255,255,255")`, and `HEX(#FFFFFF)` are the same.

## Additional Info:


cg-223:
isn't there an RGB function that serves that second purpose? having HEX also take a different format might be confusing

Aurelius7309:
the base HEX function doesn't take a leading "#", that's not something that should be changed. Different string formats being accepted by the same function also sounds confusing like cg pointed out. Finally there's not much use to the table form either as you can just make a colour table directly at that point. That said I don't see much use in this PR

english5040:
I think we can deny this PR, I agree there's not really a usability benefit.

PR #910, opened 2025-08-22
add `create_card` function to cards

fey-fish:
cards can now be declared with a `create_card` function that is called whenever said card is created using SMODS.create_card.

I found a need for it when I wanted a config variable to be a pseudorandom_element but didnt want to use add_to_deck()

## Additional Info:


Eremel:
You should used the `set_ability` function for anything that you need to do when the card is created.

PR #937, opened 2025-09-01
Immutable Card API

the-Astra:
API that allows cards to not be modified via set_ability, set_seal, set_base, copy_card, etc. Works similarly to `SMODS.is_eternal` where it runs a context `context.check_immutable` and returning `immutable = true` makes the checked card immutable. The functions that now incorporate `SMODS.is_immutable` also have a new argument `override_immutability` where cards will always be mutable no matter what if passed as `true`.

Potential issues:
- Game States where cards are created could potentially pose an issue but I have not run into anything glaring through testing. Could potentially be addressed by adding a `trigger` argument like how `SMODS.is_eternal` does
- Overriding is currently only possible by passing in the boolean argument for the functions, not by objects returning `immutable = false`

## Additional Info:

### DIFF
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -1801,63 +1801,141 @@ end
 -- OR another card's self.edition table
 -- immediate = boolean value
 -- silent = boolean value
-function Card:set_edition(edition, immediate, silent, delay)
-	SMODS.enh_cache:write(self, nil)
-	
-	if self.edition then
-		self.ability.card_limit = self.ability.card_limit - (self.edition.card_limit or 0)
-		self.ability.extra_slots_used = self.ability.extra_slots_used - (self.edition.extra_slots_used or 0)
-		if self.area then self.area:handle_card_limit(-1 * (self.edition.card_limit or 0), -1 * (self.edition.extra_slots_used or 0)) end
-	end
-
-	local old_edition = self.edition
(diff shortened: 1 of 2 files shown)


english5040:
This is a large change and I'm worried about the ramifications. I question if this is really general enough to be necessary.

I guess, a question for you: do you think there's a way to get the behavior you want with smaller changes or components, or in a more elegant way? (this is not a test, just curiosity).

philosophical side note: I am worried about effects that do something + prevention effects getting into an "arms race". not an argument I'm considering strongly

the-Astra:
> This is a large change and I'm worried about the ramifications. I question if this is really general enough to be necessary.
> 
> I guess, a question for you: do you think there's a way to get the behavior you want with smaller changes or components, or in a more elegant way? (this is not a test, just curiosity).
> 
> philosophical side note: I am worried about effects that do something + prevention effects getting into an "arms race". not an argument I'm considering strongly

It's definitely a bit more of a niche feature, however I have actually spoken with a few others who have shown support for something like this. However, due to the sheer number of ways a card object can be changed, I don't particularly see an implementation with reduced changes being super viable, though I would not mind being proven wrong. As for the "arms race" you mention, due to its niche nature, I don't believe that should be too much of an issue since most devs probably won't interact with it in the first place.

PR #943, opened 2025-09-04
Small tweak and expansion to Operator API

HuyTheKiller:
## Commit explanation:
* 1st commit: The way `SMODS.scoring_parameter` injects itself into `update_hand_text` doesn't remove `chips` and `mult`'s vanilla behavior, so I'm excluding those from the loop. I've tested and they should work without further problems.
* 2nd commit: The API injection at runtime makes some of the elements in `G.hand_text_area` refer to orphaned tables (Object nodes that no longer store an object due to it being overridden), therefore I'm making a new function that refreshes `chips` and `mult` as well as adding new scoring parameters if applicable.
* 3rd commit: Back to 1st commit, the previous injection makes `mult` behave inconsistently with the juice effect. It is restored now, but I feel like this could be expanded, so I'm adding a new argument for the `vals` table in `update_hand_text`: `parameter_juice` - which shall apply the juice effect previously seen in `mult` when set to `true`. This could easily be achieved within the `modify` method of a scoring parameter. 2nd commit also addresses `G.hand_text_area` having non-existent elements for new scoring parameters.

## Additional Info:

### DIFF
--- a/lovely/scoring_calculation.toml
+++ b/lovely/scoring_calculation.toml
@@ -129,6 +129,19 @@ for name, parameter in pairs(SMODS.Scoring_Parameters) do
 end
 '''
 
+# pt2
+[[patches]]
+[patches.pattern]
+target = 'functions/common_events.lua'
+match_indent = true
+position = 'before'
+pattern = '''
+if vals.chips and G.GAME.current_round.current_hand.chips ~= vals.chips then
+'''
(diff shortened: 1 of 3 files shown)


english5040:
I haven't checked out the Operator API but the changes seem logical and good. Thanks for doing testing

PR #945, opened 2025-09-05
Fixed infinite "Getter" context loops

AllUniversal:
*Title, what I mean by "Getter" context is any of these contexts (currently), `mod_probability`, `fix_probability` and `check_enhancement` (and `get_ranks` if Quantum Ranks gets merged)

This fix checks the following before `eval`ing any object (currently only during the `SMODS.calculate_card_areas` of a complete `SMODS.calculate_context` call, no external `eval` calls).
(The example uses Quantum Rank stuff but this problem occurs with just Quantum Enhancements too)

Example:
Joker A: All lucky cards count as 7s
Joker B: All 7s count as lucky cards
Joker A is eval'd in a `get_ranks` context, calls `SMODS.has_enhancement()` and causes a `check_enhancements` context
Joker B is then eval'd, calls `Card:is_rank()` and causes a `get_ranks` context
The game gets stuck in an infinite loop and burns into a pile of ashes

Fix:
Joker A is eval'd in a `get_ranks` context, causes a `check_enhancements` context and is set as that context's caller in the `SMODS.context_stack`.
Joker B is then eval'd and causes a `get_ranks` context.
`SMODS.calculate_card_areas` checks _before_ calling `eval_card()` on Joker A whether a previous _getter_ context of the same type as the current getter context (`get_ranks`) in the `context_stack` has caused this object (Joker A) to call any getter context, if yes, it skips evaluating Joker A.
No loop is caused! :) 

In simpler terms, this prevents an object from directly or indirectly calling itself through getter contexts.

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -288,7 +288,7 @@ if card.area and area_set[card.area] then
     if type(jokers) ~= 'table' then jokers = nil end
     if jokers or triggered then
         ret.jokers = jokers
-        if not (context.retrigger_joker_check or context.retrigger_joker) and not (jokers and jokers.no_retrigger) and not context.mod_probability and not context.fix_probability then
+        if not (context.retrigger_joker_check or context.retrigger_joker) and not (jokers and jokers.no_retrigger) and not SMODS.is_getter_context(context) then
             local retriggers = SMODS.calculate_retriggers(card, context, ret)
             if next(retriggers) then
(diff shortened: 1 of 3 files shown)


Aurelius7309 (changes requested):
Looks good though I should test this on some use cases before merging. This one's also missing LSP defs though, would be nice to have those added

Aurelius7309 (approved):

PR #947, opened 2025-09-05
Add `is_internal` argument to SMODS.destroy_cards

Yule42:
Adds a new argument `is_internal` to SMODS.destroy_cards that marks the card as being destroyed internally as opposed to externally for modded Jokers.

## Additional Info:


english5040:
I'm generally suspicious about adding on another "point of agreement" that needs to be handled by each mod. I'd like this much more if it could be made automatic without every mod having to deal with it.

I really get where you're coming from, it's important to have these points of agreement in the API, and I like the PR idea on these grounds. It's also even more important to get it right in these cases, though.

(advising not to merge currently)

Less relevant beefs:
API generality: This `internal_destruction` field is not general enough. Maybe a "reason" for destruction, so that mods can tell what caused the destruction- right now it's binary
code style: the function has too many parameters, it's getting too long. we can just use an "args" table at the end

PR #953, opened 2025-09-09
Shop Card Editing 

the-Astra:
This PR adds two new contexts: `create_shop_card` and `modify_shop_card`

- `create_shop_card` is run when a shop card is about to be created and specifies the card's `type`. Returning any of the following values will change the attributes of the shop card that will be generated:
   - `type` (string): Changes the type of the card that will be generated
   - `rarity` (number): Allows the rarity of the card that will be generated to be fixed
   - `legendary` (boolean): Returning `true` will guarantee the generated card will be legendary
   - `soulable` (boolean): Returning `true` will allow soul cards of the card type to spawn in the shop (same odds as a booster)
   - `forced_key` (string): Allows exact specification of what card will spawn

- `modify_shop_card` is run after a shop card is created and specifies the `card` object itself

## Additional Info:

### DIFF
--- a/lovely/shop.toml
+++ b/lovely/shop.toml
@@ -251,3 +251,19 @@ card:set_edition(edition)
 position = 'at'
 match_indent = true
 payload = '''card:set_edition(poll_edition('illusion', nil, true, true))'''
+
+
+# modify_shop_card context for allowing shop cards to be modified (duh)
+[[patches]]
+[patches.pattern]
+target = 'functions/UI_definitions.lua'
+pattern = '''
+local card = create_card(v.type, area, nil, nil, nil, nil, nil, 'sho')
+'''
+position = 'at'
+payload = '''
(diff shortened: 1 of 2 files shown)


english5040:
This looks very reasonable in terms of proposed functionality. Slightly worried about naming again w/ the flags as their names are very general ("rarity", "legendary", etc.).

Maybe putting them underneath one table called create_flags or similar? so it only uses up one name. then you have `create_flags.rarity`, etc.

the-Astra:
> This looks very reasonable in terms of proposed functionality. Slightly worried about naming again w/ the flags as their names are very general ("rarity", "legendary", etc.).
> 
> 
> 
> Maybe putting them underneath one table called create_flags or similar? so it only uses up one name. then you have `create_flags.rarity`, etc.

No clue why I didn't think of that, it's a much more elegant solution. I'll probably make it use this sometime tomorrow

PR #583, opened 2025-03-24
Add support for animations to centers, backs and floating sprites

Aurelius7309:
This currently doesn't support front sprites. Animated backs currently crash when using Galdur
## Additional Info:

- [ ] I didn't modify api's or I've made a PR to the [wiki repo](https://github.com/Steamodded/wiki).
- [ ] I didn't modify api's or I've updated lsp definitions.
- [x] I didn't make new lovely files or all new lovely files have appropriate priority.

PR #982, opened 2025-09-21
Added `SMODS.create_sprite()` to decide between `Sprite` and `AnimatedSprite` depending on the Atlas

AllUniversal:
+Title

With this PR the game now supports `AnimatedSprites` mostly everywhere. (I think. I should test it. Maybe :) )
`SMODS.get_atlas()` and `SMODS.create_sprite()` are two new functions to help with getting Atlases and creating `Sprites` or `AnimatedSprites` accordingly.

! The scope of the lovely patches I added maybe should be reduced, not all the targets _need_ animation compatibility + some patches don't need to use regex. (And there's at least one case where `ASSET_ATLAS` is still used directly) (`functions/UI_definitions.lua:~145 create_UIBox_notify_alert()`)

## Additional Info:

### DIFF
--- a/lsp_def/utils.lua
+++ b/lsp_def/utils.lua
@@ -752,7 +752,7 @@ function SMODS.get_atlas_sprite_class(atlas_key) end
 
 ---@param ... any The same parameters as Sprite() takes individually. The atlas may be an atlas_key instead.
 --- This function creates a Sprite or AnimatedSprite depending on the atlas passed
-function SMODS.create_sprite(...) end
+function SMODS.create_sprite(X, Y, W, H, atlas, pos) end
 
 ---@param key string The key or name of the Blind to check
 ---@param ignore_disabled? boolean Whether to ignore the Blind being disabled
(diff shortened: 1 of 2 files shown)


english5040:
Yeah this much patching is unsettling.

Hm, if it really is the case that each vanilla atlas has a unique key and is placed in exactly one of G.ASSET_ATLAS and G.ANIMATION_ATLAS then you could use a metatable to catch that access and you'd be able to route it properly. Then Sprite()/AnimatedSprite() can be overridden to work?

Got any other ideas?

AllUniversal:
I think this also closes #583 btw.

Aurelius7309:
Yep, it's also just my unfinished work and not at all in a functional state. In fact it's so stale it should have already been closed or replaced by now

english5040:
Looks like we're just leaving the patches as they are, despite the discussion on how to simplify it. We can leave it if it's more trouble than it's worth.

AllUniversal:
Oh I'll take a look at whether all the patches are needed now. Am I forgetting a way to simplify them which we had discussed? Because it's already a single kind of regex (for multiple files) and a handful of pattern patches.

PR #983, opened 2025-09-22
Better sticker apply processes

the-Astra:
Allows better functionality when applying and removing stickers. Runs two new contexts `apply_sticker = true` and `remove_sticker = true` along with the card and sticker. 
- `apply_sticker` context also differentiates between stickers applied to cards by the game (i.e. higher stake shops and challenges) and stickers applied by modded objects. Done via global value `SMODS.sticker_from_game`

Also overrides vanilla sticker applying functions (e.g. Card:set_eternal) to use SMODS funcs instead to allow for backwards compatibility

## Additional Info:

### DIFF
--- a/lovely/sticker.toml
+++ b/lovely/sticker.toml
@@ -59,7 +59,9 @@ payload = '''
 for k, v in ipairs(SMODS.Sticker.obj_buffer) do
     local sticker = SMODS.Stickers[v]
     if sticker.should_apply and type(sticker.should_apply) == 'function' and sticker:should_apply(card, center, area) then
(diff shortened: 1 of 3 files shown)


english5040:
just a note that Eremel said they are going to change the sticker application API, in case you wanted to discuss

the-Astra:
> just a note that Eremel said they are going to change the sticker application API, in case you wanted to discuss

I remember hearing this at some point. I foresee these functions persisting in some shape or form when he does touch on it, hence why I made the PR now instead of waiting

PR #985, opened 2025-09-23
feat: basic zip loading

WilsontheWolf:
This is a proof of concept PR for loading mods from zip files. This is missing a few important features, and will probably rely on lovely getting support as well first.

- Mods from zips cannot be disabled (a zip can't have a .lovelyignore inserted in it)
  - A zip with a .lovelyignore in it will be treated as disabled and unable to be reanbled
  - This will likely stay broken until lovely decides how it will disable zipped mods
- SMODS cannot find itself in a zip yet
  - Will likely need lovely support to properly test.
- Lovely will likely need some different behaviour for zips with dirs in them (since otherwise every download from github source will not be loaded by lovely). SMODS currently behaves like a zip is a folder in respect to if a mod has a lovely counterpart or not.

## Additional Info:

### DIFF
--- a/libs/nativefs/nativefs.lua
+++ b/libs/nativefs/nativefs.lua
@@ -492,4 +492,52 @@ else
     end
 end
 
+
+local redirects = {}
+
+function nativefs.smodsAddRedirect(realPath, lfsPath)
+    if redirects[realPath] then return false, 'A redirect with path "' .. realPath .. '" already exists' end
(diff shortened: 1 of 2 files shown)


Aurelius7309:
With lovely 0.9.0 out, we should revisit this

WilsontheWolf:
It's being integrated in my current work.

PR #993, opened 2025-09-29
no_collection items in collection_pool & card_collection_UIBox

TheOneGoofAli:
SMODS.card_collection_UIBox & SMODS.collection_pool now have an argument option to show/gather items that use no_collection.

## Additional Info:

### DIFF
--- a/src/ui.lua
+++ b/src/ui.lua
@@ -1888,8 +1888,9 @@ SMODS.card_collection_UIBox = function(_pool, rows, args)
     args.w_mod = args.w_mod or 1
     args.h_mod = args.h_mod or 1
     args.card_scale = args.card_scale or 1
+    args.show_hidden = args.show_hidden or false
     local deck_tables = {}
-    local pool = SMODS.collection_pool(_pool)
+    local pool = SMODS.collection_pool(_pool, args.show_hidden)
 
     G.your_collection = {}
     local cards_per_page = 0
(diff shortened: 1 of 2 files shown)


Eremel:
Code looks good, but I'm not sure I understand the purpose of something like this. Do you have a specific use case in mind?

TheOneGoofAli:
Some mods may want to hide the items from the main collection pages, but not from custom-called ones could be one such use case I thought of during the initial commit.

Fulfilling the thought, perhaps additionally a `no_collection`-only "mode" could also be considered for those wanting to display _only_ hidden items.

english5040:
I agree with the goal, but maybe a separate function to collect stuff with ex. no_collection = true, or some other condition, would be more general.

HuyTheKiller:
Speaking of `no_collection`-only mode, I feel like we could have mods add their own hidden reason (e.g. predefined objects but locked behind dependency on another mod) as loc text.

Take a look at the following example (a streak defined as a seal in my mod):
```lua
sgt_black_streak_seal = {
    name = "Black Streak",
    text = {
        "{X:dark_edition,C:white}^^#1#{} Mult",
    },
    hidden_text = {
        "Require {C:attention}Talisman{}'s",
        "hyperoperator calculation"
    },
},
```

TheOneGoofAli:
...I accidentally discarded the commits on my fork as I used the "main" branch for the original PR... and I lost the original files for this, so, I may re-submit this PR later.

PR #994, opened 2025-09-29
add a more robust way to upgrade base hand score

MetaNite64:
This PR overhauls the `level_up_hand` function defined for Scoring Parameters into a more general `upgrade_hand` function. In particular, it adds a new optional `behavior` argument: if no `behavior` is passed, the function now adds `amount` directly to the parameter's base value for the given poker hand.

`behavior` can be the strings "levelup" or "overwrite". "levelup" behaves similar to how `level_up_hand` functioned before, but it no longer overwrites bonuses given to the parameter outside of hand levels. "overwrite" directly sets the parameter's base value for the given poker hand to `amount`.

Additionally, `behavior` can also be a custom function. This function should take two arguments (which will be the current value of the parameter for the given poker hand and `amount`, in that order), and return the desired value of the parameter for the given poker hand.

Beyond the function itself, there's one other change here that only serves to make sure the vanilla `level_up_hand` function continues behaving as expected.

## Additional Info:

(the wiki doesn't have a Scoring_Parameter page yet :3)


english5040:
After thinking about this I unfortunately have to disapprove the PR. If your scoring parameter should be leveled up differently, you can simply override the level_up_hand function for it.

If the usual case is that a scoring parameter is independent of chips/mult, then it's a matter of defaults: the default level_up_hand can be for chips/mult only, otherwise a no-op.

MetaNite64:
Fair enough; I will note that this primarily targets methods to upgrade scoring parameters (including vanilla chips/mult) individually, instead of just through a standard level up. A good example of this is the Superplanets in the Vallkarri mod, which have effects like "^1.1 Mult to all hands" (they're inspired by similar planet cards from Polterworx if I remember correctly).

In any case, this PR also included a tweak to standard level up behavior to allow bonuses like the aforementioned to persist through standard level ups; currently, leveling up a hand forcibly re-calculates what its value would be without any of these extra bonuses, which is I believe different behavior than in the past. Would a smaller PR that just tweaks that be allowed through?

Eremel:
I have to disagree with english5040 here. This PR adds needed functionality - allowing for permanent modifications to hand scoring outside of levels is a feature that I have plans to implement in the future (I've been so busy recently that getting chance to work on new features has been minimal). I think that building the functionality into the parameter level up function makes a lot of sense, and having the option for different behaviours in different scenarios makes sense to me. I am definitely against changing the name of the function - it's a breaking change for no reason. Your PR also removes vanilla functionality of having a lower limit of 0 for the values - this needs reimplementing before this can be merged. There should also be some util functions provided for upgrading in these alternate ways to negate the necessity of writing a loop through all parameters yourself.

I'm happy for you to address these yourself if you want to, or I can address them for you when I sit down to do a proper session of merging some PRs again (should be soon 🤞)

MetaNite64:
Thanks! I can revert the function name change and reimplement the lower limit soon-ish, but I think you have a better idea of how you want the util functions to work so I'll leave that up to you if you don't mind.

english5040:
I don't think I was clear- "permanent modifications to hand scoring outside of levels" is really important, I just take issue with how it's done in the PR. I propose level_up_hand staying as is and another upgrade_hand function that uses the PR's "custom function" behavior; permanent modifications get an internal variable of some kind so level_up_hand respects permanent mods

apologies. I should have explained what specific ideas in the PR I like/don't like

english5040:
I propose using the "custom function" behavior only, instead of the other behaviors. Writing "X = X + 1" in your function shouldn't be too hard and looks clearer to me

Eremel:
`level_up_hand` needs to change regardless, recalculating the effective totals from scratch every time the hand is levelled up needlessly limits potential modifications you can do, for example, altering the gain rate for future level ups.

Having the custom function be called through `level_up_hand` or `upgrade_hand` (or whatever you want to call it) makes no difference because this PR should come along with a suite of util functions to simplify this process and access these internal functions for you.

I could see the `behaviour` *(please British spelling for the sake of consistency)* parameter being explicitly a function though, the `level_up_hand` function can provide a function rather than a string here to be used here.

english5040:
er, I'm brainstorming what functions we might want without being too specific. I won't discuss further

MetaNite64:
sorry for the delay, I've made the requested adjustments (lower limit of 0, reverted the function name change, and made `behaviour` spelled the British way). If/when brainstorming for the utility functions comes up again, feel free to ping me in discord!

PR #996, opened 2025-10-02
Fix challenge unlock counter

wingedcatgirl:
The challenge list still assumed challenges unlocked as in vanilla; made it actually go and count unlocked challenges instead.

## Additional Info:


english5040:
A small request: make a function to check if a challenge is unlocked, so we're not making a big if statement in multiple places

PR #997, opened 2025-10-02
Card.selectable_from_pack null check

blamperer:
I'm not sure exactly what error I had gotten that this fixes (It may have been a voucher pack from the Betmma Mods suite crashing?), but this _does_ fix it.

## Additional Info:


english5040:
Actually I think this check is needed now for mod compat: probably selectable_from_pack was called by means other than smods opening a pack under smods' control

PR #1003, opened 2025-10-05
add text_alignment to Scoring_Parameter

### LINKED ISSUE #984: Level up Chips and Mult +X numbers are center-aligned when using e.g. Burnt Joker, different from vanilla
This is due to the recent Scoring_Parameter commit; looks like `cover_align = 'cm'` is different from vanilla chips/mult `cover_align = 'cr'` or `'cl'`.

blamperer:
_Should_ fix #984 by adding a `text_alignment` property to SMODS.Scoring_Parameter.

## Additional Info:

### DIFF
--- a/lsp_def/classes/scoring_calculation.lua
+++ b/lsp_def/classes/scoring_calculation.lua
@@ -61,6 +61,7 @@ SMODS.calculate_round_score = function(flames) end
 ---@field key string Used to reference the parameter, mod_prefix is added
 ---@field default_value number Default value of the parameter
 ---@field colour? table HEX colour
+---@field text_alignment? string Must be two letters, first indicates vertical alignment, second indicates horizontal alignment
 ---@field calculation_keys? string[] Valid return keys from calculate functions
 ---@field hands? table[] Used to add custom values for different poker hands
 ---@field modify? fun(self: SMODS.Scoring_Parameter, amount: number) Alters the value of the parameter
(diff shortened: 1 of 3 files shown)


english5040:
Asking for a name change here: text_alignment implies it's for the normal text, or for all text. It actually applies to only the pop-up "+X" text that appears. Maybe "pop_up_text_alignment". (Also in future: the pop-up text alignment should probably default to the ordinary text alignment)

Functionality works great, thanks for the PR.

Seems like the ScoringParameter functionality and Chips/Mult UI are pretty disconnected (SMODS.GUI.chips_container is hardcoded), smods will need to deal with that in the future.

lsp_def/classes/scoring_calculation.lua
```diff
@@ -61,6 +61,7 @@ SMODS.calculate_round_score = function(flames) end
 ---@field key string Used to reference the parameter, mod_prefix is added
 ---@field default_value number Default value of the parameter
 ---@field colour? table HEX colour
+---@field text_alignment? string Must be two letters, first indicates vertical alignment, second indicates horizontal alignment
```
english5040, line 64: name change as per comment

english5040 (changes requested):
.

english5040:
sorry, closing due to discussion over in smods discord. a different fix is needed

PR #1006, opened 2025-10-07
Add shader patching for lovely 0.9.0+

cg-223:
This feature is reliant on the [lovely 0.9.0 apply_patches](https://github.com/ethangreen-dev/lovely-injector/pull/300) function, but does not create backwards incompatibility; 
If apply_patches does not exist, it does nothing.

## Additional Info:

### DIFF
--- a/lovely/shaders.toml
+++ b/lovely/shaders.toml
@@ -13,7 +13,7 @@ payload = '''
 local shader = "resources/shaders/"..filename
 local lovely_success, lovely = pcall(require, "lovely")
 if lovely_success and lovely.apply_patches then
-    assert(lovely.apply_patches(filename, love.filesystem.read(shader)))
+    shader = assert(lovely.apply_patches(filename, love.filesystem.read(shader)))
 end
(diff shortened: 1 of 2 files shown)


lovely/shaders.toml
```diff
@@ -0,0 +1,36 @@
...
+position = "at"
+payload = '''
+local shader = "resources/shaders/"..filename
+local lovely = require"lovely"
+if lovely.apply_patches then
+    local res, err = lovely.apply_patches(filename, love.filesystem.read("resources/shaders/"..filename))
+    if res then
+        shader = res
+    else
+        sendErrorMessage(err, 'ShaderPatcher')
```
WilsontheWolf: It may be worth the error case just throwing (assert the call to apply_patches). I don't really think it's expected behaviour to silently use the original file if lovely fails (in lua buffers this would panic)

WilsontheWolf (changes requested):

cg-223:
and i screwed up with the squash again. How do I do this every time

unscrewed up the squash :)

PR #1007, opened 2025-10-07
Fix Orbital Tag missing poker hand, expand `add_tag` function

HuyTheKiller:
~~A wrapper function to `add_tag` that initializes Orbital Tag's Poker hand if there isn't one. Also accepts a table input, much like `SMODS.create_card`~~ This PR aims to fix Orbital Tag not initializing poker hand when added at an arbitrary time (e.g. Boss Blind cashout screen, shop after Boss Blind). Also a hook to `add_tag` allows it to accept a table input.

## Additional Info:

### DIFF
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -2418,7 +2418,7 @@ function add_tag(_tag)
 	_tag.key = _tag.key or 'unknown'
 	assert(G.P_TAGS[_tag.key], ("Could not find tag \"%s\"."):format(_tag.key))
 	if not (_tag.is and _tag:is(Tag)) then
-		_tag = Tag(_tag.key, nil, _tag.blind_type or 'Small')
+		_tag = Tag(_tag.key, nil, _tag.blind_type)
 	end
 	add_tag_ref(_tag)
 end


HuyTheKiller:
I tried to make it as concise as possible. Would "fresh tags" fit the description of "not copying a tag"?

Eremel:
I think this should just be patched into either tag initialisation or into the `set_ability` function of tag objects, there's no need for another utility function that would require everyone transitioning to using here imo

HuyTheKiller:
I thought of that at first too, but for some reason I chose to have another util function instead 🤦
Your approach includes patching Orbital Tag's `set_ability` and overriding `add_tag`, am I correct?

Eremel:
No, just patching either `Tag:init` or `Tag:set_ability` - both of these functions are **always** called whenever a tag is created (during a run), no matter what you do with it afterwards. I'd prefer set ability due to it not being used for collection tags.

HuyTheKiller:
There. Orbital Tag fix moved to patch.
`add_tag` hook allows it to accept a table input. Example:
image

english5040:
uh, yeah that does look better. sorry

english5040:
Alright I'm calling it good, thanks!

PR #1014, opened 2025-10-12
Allow calc_scaling to apply to editions

jumbocarrot0:
Adds a small bit of code to `SMODS.scale_card` that checks if any cards have an edition with the `calc_scaling` function. This allows editions to modify the scaling effect of cards like jokers can.

## Additional Info:


english5040:
alright, agreed that this is a good addition.

I'm concerned (outside of this specific PR) that we have many different things happening in calc, each of which is passed to an inconsistent subset of objects. Better Calc is good at passing things to every relevant cardarea but there still are gaps

PR #1026, opened 2025-10-17
Fixes for hand level-up animations with other scoring parameters

Oinite12:
Currently, while values are properly updated during hand level-up, only Chips and Mult have the `attention_text` overlay during the hand level-up animation. Furthermore, said overlays are improperly sized when a scoring calculation with `replace_ui` is set. This PR seeks to fix these issues.

Hand level-up animations are now performed with functions that play one of three stages of the hand level-up animation, that being:
- initialization (display poker hand information),
- iteration (flash the `attention_text` overlay one by one), and
- resetting (back to default scoring parameter values).

These stages need to be separated since `level_up_hand` only plays the iteration stage, and not initialization and resetting (which is instead done by whatever item is leveling up the hand), except for Black Hole, which manually plays all three stages (since it has `level_up_hand -- instant=true`).

To make use of the animation stage functions, Black Hole, Burnt Joker, and Orbital Tag have their ownership taken since it's a bit more simpler to do so than patching out entire blocks of code. `level_up_hand` is overridden for a similar reason.

A note on iteration: The order that the scoring parameters take to have an `attention_text` overlay appear is dependent on the order of keys in `G.GAME.current_scoring_calculation.parameters` (i.e. the `parameters` table of a scoring calculation), thus the default `parameters` table in `SMODS.Scoring_Calculation` was also changed to switch the order of "mult" and "chips".

## Additional Info:

### DIFF
--- a/lovely/scoring_calculation.toml
+++ b/lovely/scoring_calculation.toml
@@ -332,21 +332,6 @@ payload = '''
 SMODS.Scoring_Parameters.chips:modify(nil, _chips - (hand_chips or 0))
 '''
 
-# Properly target SMODS-defined chips/mult text area for level up hand animation
-[[patches]]
-[patches.pattern]
-target = 'game.lua'
-match_indent = true
-position = 'at'
-pattern = '''
-chips = self.HUD:get_UIE_by_ID('hand_chips'),
-mult = self.HUD:get_UIE_by_ID('hand_mult'),
-'''
-payload = '''
-chips = self.HUD:get_UIE_by_ID('hand_chips_area'),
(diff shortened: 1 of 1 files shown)


Oinite12:
Regarding previous commit, I forgot to mention that it's a temporary and visual blanking that I fixed

Oinite12:
I realized there were some changes here that's causing some additional animation bugs; in particular, when the text changes in the mult score container, the _container_ juices up, not the text itself. I made a grave error when debugging for this PR by giving the container of the score container the ID `"hand"..name` (following the [1.0.0-BETA-0827 example](//github.com/Steamodded/smods/discussions/919)), which is supposed to be (and is) used by the score container text.

PR #1035, opened 2025-10-22
Add soft reload keybind

TheOneGoofAli:
Since discovering the ability to reload the localization file without restarting the game, I figured it'd prove to be a useful enough feature in SMODS itself as a keybind - holding "L" (for as much as holding "M") now kicks off such a reload.

Edit: this also reloads any changes done for existing atlas files, which may prove useful for those wanting to see how the changes for a particular item look in-game, but this does not reflect location changes of the object on said atlas.

## Additional Info:

### DIFF
--- a/localization/en-us.lua
+++ b/localization/en-us.lua
@@ -165,10 +165,28 @@ return {
                     '{C:dark_edition}#1#{} area slots'
                 }
             },
+            generic_card_limit_pc = {
+                name = "Hand Size",
+                text = {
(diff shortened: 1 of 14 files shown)


TheOneGoofAli:
Changed title and amended initial message, this very much also affects changes on atlases and such.

Eremel:
After discussing with @WilsontheWolf, we think a button like this would be better suited for Debug Plus.

PR #1038, opened 2025-10-25
Overhaul SMODS.destroy_cards()

english5040:
- Returns destroyed cards in `.destroyed`
- uses a table for arguments
- skip_destroy_animation is refactored out
- rename `skip_anim` arg. to `pinch_anim`, this makes more sense as we are using a specific animation
- `args.delay` inserts a delay between destroy animation and calc. Vanilla Hanged Man uses 0.3. In the general case, this makes the result look a lot better
- `args.no_run` allows the user to use the logic of this function however they want (see Immolate for use case)

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -1810,25 +1810,6 @@ payload = '''
         local flags = SMODS.calculate_context({joker_type_destroyed = true, card = self})
         if flags.no_destroy then self.getting_sliced = nil; return end
     end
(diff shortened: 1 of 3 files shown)


AllUniversal:
Can this be revived or should I make my own PR?

english5040:
Apologies, I'm not very active currently. go ahead and make your own if that's easier

nh6574:
Implemented by #1401

PR #1056, opened 2025-11-01
Fix crash if a modded stake has multiple applied stakes

wingedcatgirl:
`SMODS.applied_stakes_UI()` has never returned anything, so attempting to assign its result to `num_added.val` resulted in niling it out and crashing if the game looked at it again in the same loop. Which conveniently only happened if people gave their stakes a branch-and-bottleneck structure that I guess nobody ever thought was worth the effort. 

It's fucky in some other ways but that's the one I think I understand enough to fix.

## Additional Info:


wingedcatgirl:
Closing this cause it's included in 1058 now.

PR #1059, opened 2025-11-01
Show the Cat during loading

TheOneGoofAli:
image

If the 96x96 "cat.png" image exists in "assets", it will be drawn on the loading screen with a bounding rectangle much like the progress bar... and if it doesn't, it is simply not drawn.

Originally [done](https://github.com/TheOneGoofAli/TOGAPackBalatro/commit/5f6a7593304f48be20dea7c7fbf93a226c7215c5#diff-300651834a6b7ea4e4f53e418649987f3b595a21713bc61afe4e40be38a98cc5) for my own mod, whilst the implementation for it uses base64 encoding/decoding for storing the image, for this PR, I simply added the actual .png file for direct use instead.

> ...makes me think of the Windows 8.x login screen to an extent.

## Additional Info:


the-Astra:
I mentioned this in balacord but I'm saying it here again for visibility. I think it would be really cool if this also showed the modicon.png of whatever mod was currently being loaded

WilsontheWolf:
I am in the process of completely reworking the boot process including the boot screen and my chances will end up overwiriting any changes. I'm not sure if I think this is a good idea or not, I'll have to ponder it.

TheOneGoofAli:
If all else, could resubmit the PR with said changes in mind later. 😅

PR #1062, opened 2025-11-02
Allow shaders to be applied to UI boxes

InvalidOS:
This adds a lovely patch that allows for shaders to be applied any UI element that is drawn using `UIElement:draw_pixellated_rect()`.

In order to apply a shader to a UI element, you must add a new config parameter, `shader`, to the UI node you want to apply a shader to. This may be structured in a few ways:

- `shader = "shader_key"`, which will simply use the shader `SMODS.Shaders["shader_key"]` with default arguments being sent.
- `shader = { shader = "shader_key", send = tbl}`, which uses `SMODS.Shaders["shader_key"]`, while sending arguments defined by `tbl` (this works the same way as the `_send` argument in `Sprite:draw_shader()`)
- A table of tables formatted either of the above two ways, which will result in each one being drawn in order, akin to how `Sprite:define_draw_steps()` functions. Using the string `"none"` or `"dissolve"` as the shader key will instead result in no shader being used, which is useful for layered effects.

UI element shaders, by default, use a different parameter set than the typical sprite shaders, so they must be custom-made to work with this. The only arguments shared between them are `screen_scale`, `time` (which uses a placeholder value), and the one using the shader's key. UI element shaders additionally send both the position and size of the UI element as `uibox_pos` and `uibox_size`, respectively.

To see an example of this in action, [Entropy](https://github.com/lord-ruby/Entropy) currently has an implementation of UI element shaders I made earlier (which I modified to create this PR), used to apply a shader to its mod badge. [Arrow API](https://github.com/Kekulism/Arrow) also has a separate implementation to apply the Polychrome shader to UI elements.

I may come back to this to add support for text nodes and DynaText objects, but I'm already very happy with this.

## Additional Info:


lovely/ui_elements.toml
```diff
@@ -149,3 +149,91 @@ if deck_center.check_for_unlock and type(deck_center.check_for_unlock) == "funct
...
+            G.TIMERS.REAL/28,
+            G.TIMERS.REAL
+        }
+        
+        if shader == "none" or shader == "dissolve" then
+            _draw()
+            return
+        elseif send then
+            for _, v in ipairs(send) do
+                local val = v.val or (v.func and v.func(self))
```
stupxd: `v.ref_table[v.ref_value]` would be nice to have for consistency

InvalidOS:
I attempted to get a different branch working using CanvasSprite but that ended up being too difficult for me to get working

InvalidOS:
converting this to a draft again to add some changes by @lord-ruby

InvalidOS:
didn't say this earlier but this should be ready to be merged

PR #1067, opened 2025-11-08
Add context.blueprint_copier

InvalidOS:
This PR adds `context.blueprint_copier`, which contains the Joker in the `copier` argument of `SMODS.blueprint_effect`. This also lovely patches Blueprint and Brainstorm to use `SMODS.blueprint_effect` so they also use this context.
This is useful for multi-blueprint effects trying to avoid infinite loop interactions, as they can check if they're attempting to copy `context.blueprint_copier` and not copy the card if they are. Two-way blueprint effects are quite common, so many of them would find this useful.

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -2098,14 +2098,17 @@ function SMODS.blueprint_effect(copier, copied_card, context)
     local old_context_blueprint_card = context.blueprint_card
     context.blueprint_card = context.blueprint_card or copier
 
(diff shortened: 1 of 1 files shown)


english5040:
just noting here: it would make sense to have a list/chain of copiers, otherwise a joker couldn't check if it's copying itself after a chain of copies like A->B->...->A

InvalidOS:
the pr merge here was just me resolving merge conflicts

PR #1072, opened 2025-11-11
Allow passing a number as `scalar_value` in SMODS.scale_card

wingedcatgirl:
For cases where you want to scale by something trivial like 1

## Additional Info:


Eremel:
This encourages bad coding practices, along with causing issues for any mods that want to manipulate the scaling values in a permanent way.

Regardless of that, any numbers that are used should be stored in the config of an object anyway, so the function only handles these to promote correct practices.

PR #1074, opened 2025-11-13
Add optional automatic generation for atlases in centers

cg-223:
This PR allows you to write code like the following:

```lua
SMODS.Joker { --or any other center
    key = "my_joker",
    path = "my_joker_image.png",
}
```
omitting the `SMODS.Atlas` entirely. The above code is equivalent to this:
```lua
SMODS.Atlas {
    key = "my_joker_atlas",
    path = "my_joker_image.png",
    px = 71,
    py = 95
}

SMODS.Joker { 
    key = "my_joker",
    atlas = "my_joker_atlas"
}
```

SMODS creates an automatically generated atlas pointing at `assets/xx/my_joker.png`. This is intended to be used for atlases that contain an image for just one center. As such, a custom `pos` is not supported (but `soul_pos` is).

`SMODS.Center.register` handles the atlas generation, but `default_size` (which is what informs `SMODS.Center` what should be filled in for `px` and `py`) is optional for extending classes (despite all 7 centers using 71x95 by default). This is because some mods extend `SMODS.Center` and have a different default size (take `CardSleeve`, which is 73x95).

Setting `size` as a parameter overrides `default_size`, and setting `px` and `py` individually also overrides the size used. Other than that, you can't use other parameters for `SMODS.Atlas` in `SMODS.Center`s.

Encouraging mods to not have monolith asset files is nice, mainly for collaborative purposes. Conflicts in asset files are not easy to resolve, and often require a lot of manual labor. When multiple people are working on the same mod at once, having one file per card eliminates the risk of needing to do all this tedious work. 

Also: I'm not sure how best to describe this on the wiki, should there be a note in `SMODS.Atlas` stating you can do this? I'm sure the `path` parameter should be described on the individual `SMODS.Center` pages in some way.

## Additional Info:

### DIFF
--- a/src/game_object.lua
+++ b/src/game_object.lua
@@ -33,18 +33,18 @@ function loadAPIs()
         -- also updates o.prefix_config
         SMODS.add_prefixes(self, o)
         if o:check_duplicate_key() then return end
-        if (self.default_size or (self.px and self.py) or self.size) and self.path then
+        if self ~= SMODS.Atlas and (o.default_size or (o.px and o.py) or o.size) and o.path then
             key_unique = key_unique + 1;
-            local size = self.size or self.default_size -- if a mod wants to override
-            local px, py = self.px or size.px, self.py or size.py
+            local size = o.size or o.default_size -- if a mod wants to override
+            local px, py = o.px or size.px, o.py or size.py
             SMODS.Atlas {
-                key = "__autogen_atlas_" .. self.key .. "_" .. key_unique,
-                path = self.path,
+                key = "__autogen_atlas_" .. o.key .. "_" .. key_unique,
+                path = o.path,
                 px = px,
                 py = py,
                 raw_key = true,
             }
-            self.atlas = "__autogen_atlas_" .. self.key .. "_" .. key_unique
+            o.atlas = "__autogen_atlas_" .. o.key .. "_" .. key_unique
         end
         o:register()
         return o


stupxd:
> Encouraging mods to not have monolith asset files is nice

The idea for having monolith asset files is to improve performance.

~~I wonder if it makes sense to have some SMODS logic that combines all assets into a single atlas for this case 🤔
Would that be too much?~~

~~i.e. create a list of all assets that are default size (71x95); create a canvas with optimal size that fits all assets; then draw them one by one, update pos for each center and finally create an atlas from this canvas.~~

After sleeping on it, I think my solution is completely unnecessary.
Modders can just do it manually for each release.

SleepyG11:
1. Maybe makes sense to change `path` to `atlas_path`, just for making key less generic?
2. Maybe makes sense implement same logic for `SMODS.GameObject` as a whole, not only `SMODS.Center`? This will cover objects like `SMODS.Stake`, and as a bonus custom game objects aswell.

cg-223:
@SleepyG11 
> 1. Maybe makes sense to change `path` to `atlas_path`, just for making key less generic?

I don't see how key is generic, it doesn't refer to the generated atlas.

> 2. Maybe makes sense implement same logic for `SMODS.GameObject` as a whole, not only `SMODS.Center`? This will cover objects like `SMODS.Stake`, and as a bonus custom game objects aswell.

I agree, and I've done this.



@stupxd 
> The idea for having monolith asset files is to improve performance.

Makes sense, maybe this shouldn't be used for huge projects. I can still see this being useful for small/mid sized mods, where especially for sprites that there's commonly only one or two of (enhancements, stakes, backs, etc). 

I'm not sure how big the performance hit is to split an atlas up into a dozen or two, nor do I want to spend the hour it would take to benchmark. Regardless, I think ease of development has historically been prioritized over performance in SMODS (see `calculate`) and I can't see how splitting assets into individual files would be anywhere near as bad as some of the other stuff SMODS encourages you to do.

SleepyG11:
> I don't see how key is generic, it doesn't refer to the generated atlas.

Since you're adding this to `SMODS.GameObject`, which by itself is pretty generic class, other mods may be using `path` for own needs, whatever this can be. Using more specific key decrease chance to break some mods.

english5040:
yes renaming to `atlas_path` or something is an easy improvement
On second thought: `path` implies the object itself is something akin to an atlas. should definitely change the name so it's clear the field applies to an automatically-generated, separate atlas. Perhaps fields related to the automatically generated atlas can be collected into a table? (might be awkward with lua table "inheritance")

I think this can be great for ease of development. stupxd: as long as we are not encouraging something really hurtful in the long term (which I don't think this is) I think this is ok

src/game_object.lua
```diff
@@ -32,6 +33,19 @@ function loadAPIs()
         -- also updates o.prefix_config
         SMODS.add_prefixes(self, o)
         if o:check_duplicate_key() then return end
+        if self ~= SMODS.Atlas and (o.default_size or (o.px and o.py) or o.size) and o.path then
```
english5040: this looks more like a well-formedness check. so if o.path exists, assert the presence of a default_size, px/py, or size.

I'm also uncomfortable with having both "size" and "px + py" as ways to override.

cg-223:
don't really have the energy to keep up with this actively rn

PR #1075, opened 2025-11-13
Adds tracking for hands played per ante

VivianGiacobbi:
Exactly what it says on the tin. When a hand is played in `G.FUNCS.evaluate_play()`, its ante count is updated alongside its round count, and these counts are reset during `end_round()` when a boss is defeated at the end of an ante. Additionally adds them in the definition for SMODS.PokerHand

## Additional Info:


HuyTheKiller:
I have a Showdown Blind that blocks repeat hand types this ante, and it uses exactly this feature. I've just made it yesterday.
Did you somehow read my mind?

PR #1081, opened 2025-11-15
Streamline G.viewed_stake logic

NeuralSixty:
The logic for `G.viewed_stake` can be streamlined into the following 4 cases:

1. If the profile has everything unlocked (`all_unlocked`), nothing else matters and any stake can be viewed.
2. If the stake is available, it can be viewed as well. According to the logic of `SMODS.check_applied_stakes`, a stake is considered available if it exists without depending on any applied stakes, or if it has applied stakes, all of them have to be beaten beforehand.
3. If the stake is not available, the deck usage data is checked for the most often won stake, and the stake after that is chosen.
4. If all else fails, displays the first stake overall.

An additional check to see if a stake is unlocked could be done, but I have failed to catch any edge cases after repeated manual tests. Therefore, a check for applied stakes should suffice.

This should also work for multiple stakes that could be available right from the start of a new profile (e.g. several variations of the White Stake).

## Additional Info:

### DIFF
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -572,7 +572,11 @@ function G.UIDEF.stake_option(_type)
 	local stake_options = {}
 	local curr_options = {}
 	local deck_usage = G.PROFILES[G.SETTINGS.profile].deck_usage[G.GAME.viewed_back.effect.center.key]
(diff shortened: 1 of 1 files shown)


SleepyG11:
New profile without "Unlock All"
We have 2 stakes: Red Stake (vanilla) + modded stake A, both of them require beat White Stake.
Also we have modded stake B, which require beat stake A.
We have 2 decks, deck X and deck Y, they're right next to each other.

I beat White Stake + Stake A on deck X, and then White Stake on deck Y.
I selected Stake B on deck X and switched to deck Y.

Expected behaviour: selected Stake A on deck Y
Reality: selected Red Stake on deck Y

PR #1085, opened 2025-11-18
calc_scaling operation override

real-niacat:
allows cards in calc_scaling to provide an override operation (e.g. making all other scaling jokers scale multiplicitively)

sorry for no lsp defs i didnt see anything relating to the arguments of calc_scaling's return values

## Additional Info:


stupxd:
Is it ok that multiple jokers with that effect would just override each other?

SleepyG11:
Because of #1014 maybe makes sense to make some kind of util function to apply effect of calc_scaling?

real-niacat:
i think its fine if multiple cards override each other like this, idk how you would cleanly implement it w/o having one take priority over another

PR #1086, opened 2025-11-18
Fix replace sound infinite loop

VivianGiacobbi:
The hook for `play_sound` to utilize `SMODS.Sound.replace_sounds` was incorrectly calling itself, rather than the hook ref, resulting in an infinite loop. People must not use this feature a lot since seemingly the rest of the code was written a year and a half ago and this was never caught.

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -35,7 +35,7 @@ match_indent = true
 payload = '''
 function eval_card(card, context)
     context = context or {}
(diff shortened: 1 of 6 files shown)


stupxd (changes requested):
This could be merged if this didn't include [Added SMODS.spectral_downside() calculation](https://github.com/Steamodded/smods/pull/1086/commits/5fec4bcf6d3c45708bc694f3fc65459f94484cbe), that seems like something from a different PR.

PR #1101, opened 2025-11-29
Add disable_if_empty to SMODS.Rarity

factwixard:
Whether or not it's an issue in vanilla, when a rarity is empty, only the default Jimbo spawns. This basically acts as a solution that affects specifically modded rarities.
In full: 
- Adds disable_if_empty as an optional boolean field in lsp_def/classes/rarity.lua
- A new check runs in SMODS.poll_rarity (src/utils.lua) that uses disable_if_empty, and if the value is true and the pool is empty save for the default, then the rarity will be removed from available_rarities until SMODS.poll_rarity is used again
## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -361,7 +361,7 @@ function SMODS.create_card(t)
     if not t.area and t.key and G.P_CENTERS[t.key] then
         t.area = G.P_CENTERS[t.key].consumeable and G.consumeables or G.P_CENTERS[t.key].set == 'Joker' and G.jokers
     end
-    if not t.area and not t.key and t.set and SMODS.ConsumableTypes[t.set] then
+    if not t.area and not t.key and t.set and SMODS.ConsumableTypes[t.set] or t.set == 'Consumeables' then
         t.area = G.consumeables
     end
     if not t.key and t.set == 'Playing Card' or t.set == 'Base' or t.set == 'Enhanced' or (not t.set and (t.front or t.rank or t.suit)) then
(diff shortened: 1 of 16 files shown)


english5040:
Please don't use `SMODS.remove_pool`, that isn't what the function is normally used for (it implies a permanent modification).

(The rarity behavior is indeed from vanilla.)

I think this is a fine feature to add, it's pretty short code

english5040 (changes requested):

src/utils.lua
```diff
@@ -823,6 +823,14 @@ function SMODS.poll_rarity(_pool_key, _rand_key)
     local available_rarities = copy_table(SMODS.ObjectTypes[_pool_key].rarities) -- Table containing a list of rarities and their rates
     local vanilla_rarities = {["Common"] = 1, ["Uncommon"] = 2, ["Rare"] = 3, ["Legendary"] = 4}
 
+	-- Check to see if all disabled rarities are empty
```
english5040: minor: sentence here should read "check if any rarities are empty and should be disabled"

src/utils.lua
```diff
@@ -823,6 +823,14 @@ function SMODS.poll_rarity(_pool_key, _rand_key)
...
+    for _, v in ipairs(available_rarities) do
+        local _pool = get_current_pool("Joker", v.key, false, nil)
+        if SMODS.Rarities[v.key] and SMODS.Rarities[v.key].disable_if_empty and type(SMODS.Rarities[v.key].disable_if_empty) == "boolean" and SMODS.Rarities[v.key].disable_if_empty == true and #_pool == 1 and _pool[1] == "j_joker" then
```
english5040: Actually there is a minor bug here: if the only Joker left is j_joker this would be treated as an empty pool. Modders might also want to add their own default to a rarity which would break this (we should support this also). This code might need some rethinking unfortunately.

GauntletGames-2086: Yeah, when I read the change first my immediate thought is that we should have a feature that allows rarities to have their own "default" card besides the one the ObjectType specifies. This hypothetical feature could accept both a key or something representing "none" (which would act just like this PR).

Doing such would however deprecate this PR, and require very different logic than what we're trying here.

factwixard: i do have a version of the code that supports the default field, even if it doesn't currently exist. i could add that just in case, if that's preferred

src/utils.lua
factwixard: updated wording to this.

src/utils.lua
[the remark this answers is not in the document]
english5040: sorry. `if SMODS.Rarities[v.key] and SMODS.Rarities[v.key].disable_if_empty then ... end` is fine. (removing the == "boolean" and == true)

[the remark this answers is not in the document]
factwixard: changed to just be a truthy check for the field.

factwixard:
just restructured the code to be a field check in get_current_pool which gives an override string rather than an item, and changed the part in SMODS.poll_rarity to check for the override string instead of any specific item.

PR #1105, opened 2025-12-02
Add default values for scaling functions

Perwollnt:
bad fix to nullpointer in scaling function
I found an npe when playing equal chances deck after playing some cards. This is a messy fix to only that issue, i have not checked why are the modifiers 0. This is probably a bad fix but if someone needs it or finds the same issue its here and its fast.

## Additional Info:


english5040:
Are you playing Cryptid? The bug is the mod's fault, not smods. I'd post an issue for whatever mod you're using rather than smods

PR #1106, opened 2025-12-03
consistently define "added_to_deck" for playing cards

vitellaryjr:
`Card:add_to_deck` is not called on any initial cards in your deck, nor any cards created by `create_playing_card` (used by cards such as the Certificate Joker or the Incantation Spectral), but it _is_ called for _all_ other methods of adding cards to your deck (such as taking cards from Standard Packs, or creating copies with DNA or Cryptid). this means that currently, `added_to_deck` is useless for being able to identify whether a card is in your deck or not, as it will be arbitrarily false for most (but not all) cards in your deck

this PR fixes this by calling `Card:add_to_deck` on all cards created by `card_from_control` and `create_playing_card`, making it consistent with all other methods of adding cards to your deck. from my testing, this has no gameplay consequences, and only serves to make `added_to_deck` a reliable method of identifying whether a card is currently in your deck or not

disclaimer that this is a vanilla bug, not one introduced by SMODS; i'm unsure whether fixing vanilla issues in this manner in order to make modding more convenient is something that is of interest, and so rejecting this PR on those grounds would be reasonable if such a change is undesired

## Additional Info:

### DIFF
--- a/lovely/fixes.toml
+++ b/lovely/fixes.toml
@@ -829,3 +829,15 @@ if control.g then _card:set_seal(control.g, true, true) end
 payload = '''
 _card:add_to_deck()
 '''
+## Add cards created by create_playing_card to deck
+[[patches]]
+[patches.pattern]
+target = 'functions/common_events.lua'
+match_indent = true
+position = 'after'
+pattern = '''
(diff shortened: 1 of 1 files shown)


english5040:
yes, we do fix vanilla stuff like this; this is great, thank you.

can probably be merged straight away

PR #1110, opened 2025-12-06
Add extra easings for "ease" event type

Mysthaps:
ported from EaseMore on Balacord
## Additional Info:


SleepyG11:
Just a suggestion, what if instead of multiple `elseif`, make a dictionary where all easing functions can be stored/added and accessed by a key? Cleaner + easier for other mods add own easing on demand

Mysthaps:
true, just copied old code over for this

PR #1114, opened 2025-12-12
add 'set_cost' and 'set_sell_cost' contexts

vitellaryjr:
these contexts can be used to create effects that alter the cost and sell value of cards dynamically. needs to be 2 contexts, so that the original sell value can be based on the new cost, while also allowing new sell values to be based on the original sell value. 'tis a little ugly (and will be mildly annoying to document in a concise way), but it's the best way i could find to do it and make it sufficiently flexible

## Additional Info:

  - contexts need to be documented later

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -1938,12 +1938,12 @@ self.sell_cost = math.max(1, math.floor(self.cost/2)) + (self.ability.extra_valu
 '''
 payload = '''
 if not loading then
-    local cost_flags = SMODS.calculate_context({set_cost = true, card = self}) or {}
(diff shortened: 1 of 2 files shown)


VivianGiacobbi (commented):
Similarly to other calculation keys, it might be preferred to send the card's current/expected cost in the context and then update the value in the context via `SMODS.update_context_flags()`, since currently any consequent calculation returns will simply override the value rather than modify it afaik

vitellaryjr:
aye, good call; have done that now

vitellaryjr:
(closing this due to #1310 accomplishing the same thing in a bit of a cleaner way)

PR #1126, opened 2025-12-19
Expansion of functionality provided by #994

Gloompashcy:
### **Introduces various features that expand upon the functionality provided by #994, including:**

- The ability to define the default level up behaviour of all hands and scoring parameters
- Re-implementation of the level_up_hand function in scoring parameters, allowing scoring params to have their own default behaviour
- Addition of `level_up_hand` function to Poker Hands, which has the same functionality as the same function in Scoring Parameters, but takes priority over it
- Restructuring of `SMODS.upgrade_poker_hands` to use default upgrade behaviour if no `args.func` function is defined
- `SMODS.upgrade_poker_hands` now passes `args.level_up` as a parameter into `args.func`
- New util function `SMODS.change_default_hand_upgrade_func` which changes the default behaviour for all hands/scoring params
- New util function `SMODS.get_default_hand_upgrade_func` which creates a function based on the `level_up_hand` functions from a given hand and scoring param
- This created function takes four params: `base`, the initial value of the parameter being upgraded, `hand`, the name of the hand being upgraded, `parameter`, the name of the scoring parameter being upgraded, and `amount`, the number the level is being increased by
- `SMODS.hand_upgrade_func` is the default behaviour of all poker hands and scoring parameters without a defined `level_up_hand` function; by default it is the following function:
```lua
SMODS.hand_upgrade_func = function(base, hand, parameter, amount)
    return math.max(base + G.GAME.hands[hand]['l_' .. parameter] * amount, 0)
end
```

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -929,7 +929,7 @@ pattern = '''
 position = 'at'
 match_indent = true
 payload = '''
-SMODS.calculate_context({open_booster = true, card = self})
+SMODS.calculate_context({open_booster = true, card = self, booster = booster_obj})
 '''
 
 # context.buying_card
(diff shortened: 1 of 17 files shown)


Gloompashcy:
Currently a draft because the changes to level_up_hand will probably cause compatibility problems

Gloompashcy:
Changed from returning new values to setting them to match with #994
However, the `amount` parameter passed into `SMODS.Scoring_Parameter.level_up_hand` is now the number of level ups rather than the change in the scoring param value

Issue #991, opened 2025-09-28
game creashed


ssyangchn:
```
Oops! The game crashed:
[SMODS _ "src/utils.lua"]:2143: attempt to index field 'children' (a nil value)


Development version of Steamodded detected! If you are not actively developing a mod, please try using the latest release instead.



Additional Context:
Balatro Version: 1.0.1o-FULL
Modded Version: 1.0.0~BETA-0927a-STEAMODDED
LÖVE Version: 11.5.0
Lovely Version: 0.7.1
Platform: Windows
Steamodded Mods:
    1: Quick Load by Asobi [ID: QuickLoad]
    2: JokerSellValue by OppositeWolf770 [ID: JokerSellValue, Priority: 100000, Version: 1.2.1]
    3: Better Mouse And Gamepad by Kooluve [ID: BetterMouseAndGamepad, Priority: -10000, Version: 1.0.6c]
    4: Nopeus by jenwalter666, stupxd [ID: nopeus, Version: 2.2.3, Uses Lovely]
    5: Manual Flip by ssyangchn [ID: ManualFlip, Version: 1.0.0]
    6: DebugPlus by WilsontheWolf [ID: DebugPlus, Version: 1.3.1, Uses Lovely]
    7: Flush Hotkeys by Agoraaa [ID: FlushHotkeys, Version: 1.0.4]
    8: Always Show Seed by infarctus [ID: Infarctus_Always_Show_Seed]
Lovely Mods:
    1: NoFlip
    2: DVPreview_v4.1.2

Stack Traceback
===============
(3) LÖVE metamethod at file 'boot.lua:352'
Local variables:
 errhand = Lua function '(LÖVE Function)' (defined at line 553 of chunk [lovely debugplus.console "console.lua"])
 handler = Lua function '(LÖVE Function)' (defined at line 553 of chunk [lovely debugplus.console "console.lua"])
(4) Lua field 'get_card_areas' at Steamodded file 'src/utils.lua:2143' 
Local variables:
 _type = string: "individual"
 _context = nil
 t = table: 0x0e8626f0  {1:table: 0x0e862730}
 (*temporary) = number: 2
 (*temporary) = table: 0x0df72f20  {object:"MANUAL_REPLACE"}
 (*temporary) = nil
 (*temporary) = number: 0
 (*temporary) = nil
 (*temporary) = nil
 (*temporary) = number: 0
[...]

PR #1127, opened 2025-12-19
Ensure object is non-nil in SMODS.get_card_areas

mleaf233:


mleaf233:
Related issue: https://github.com/Steamodded/smods/issues/991

PR #1135, opened 2025-12-22
`context.modify_blind_chips`

the-Astra:
This commit allows objects that trigger off contexts to change a blind's target amount and correctly shows it on the Blind UI screen (assuming the object is in the player's possession at time of UI creation). It currently calls with three attributes:
- `context.blind = table` - A reference to the Blind's config table
- `context.modify_blind_chips = true` - Base context boolean
- `context.blind_multiplier = float` - The multiplier value the blind uses to modify from the base ante amount (i.e. `1.5` for Big Blind)
- `context.blind_chips = int` - The chips value the blind holds that needs to be defeated

Returning `modify = number` in a calculate function's return table will set the blind's chips to the returned value. `context.blind_chips` is updated accordingly for subsequent triggers

## Additional Info:


Aurelius7309:
Considering #1281, is there still a use case for this?

the-Astra:
1281 looks to do what this does but better, however something I think that is missing from it is the ability to change the size text on the select screen

the-Astra:
Closed due to the merging of #1281

PR #1146, opened 2025-12-28
Add SMODS.get_clean_pool

the-Astra:
A simple helper function that returns the current specified pool without any `'UNAVAILABLE'` values. Takes the same arguments as vanilla's `get_current_pool`

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -3413,10 +3413,11 @@ end
 
 function SMODS.get_clean_pool(_type, _rarity, _legendary, _append)
     local pool = get_current_pool(_type, _rarity, _legendary, _append)
+    local clean_pool = {}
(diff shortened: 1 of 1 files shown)


nh6574:
I think it should make a new list instead of replacing all values with nil, no? I also don't think it's good practice to remove values don a table while iterating on it but I assume it works in this case.

the-Astra:
> I think it should make a new list instead of replacing all values with nil, no? I also don't think it's good practice to remove values don a table while iterating on it but I assume it works in this case.

Doing a new list is probably better. The reason I did the iteration method is because doing it in a for...in ipairs loop would cause skips on consecutive unavailable objects, but this will be moot when I make it return a new list

PR #1167, opened 2026-01-15
feat: score modification effect

Aikoyori:
Adds score modification, related calculation keys, localization (en-us only for now), perma-bonuses, `SMODS.mod_score`, and custom sound effect for xscore

Feel free to edit the PR if something looks wrong

## Additional Info:


soulware1:
i wonder if we could get eq_mult and eq_chips in too...

Aikoyori:
got it

PR #1169, opened 2026-01-17
Update LSP defs

wingedcatgirl:
tyop in `should_apply`

## Additional Info:


nh6574:
The comment should probably be updated too

PR #1174, opened 2026-01-21
add SMODS.ScreenShader

real-niacat:
## Additional Info:

adds a new gameobject - SMODS.ScreenShader, which applies a given shader to the entire screen. Any amount of screenshaders can be stacked on the screen at once without conflicting
i am not very confident that this is good but its ok :3

### DIFF
--- a/lsp_def/classes/screenshader.lua
+++ b/lsp_def/classes/screenshader.lua
@@ -30,4 +30,4 @@ SMODS.ScreenShader = setmetatable({}, {
 })
 
 ---@type table<string, SMODS.ScreenShader|table>
-SMODS.ScreenShader = {}
+SMODS.ScreenShaders = {}


ethangreen-dev:
+rep good trade

cg-223:
imo CRT should be applied later than normal since it leaves those empty bits on the corners

real-niacat:
i'll change it to order of 0 then, that feels right since it is the default shader thats always applied

Aurelius7309:
Is there a good reason for this not being a subclass of SMODS.Shader?

real-niacat:
the main reason why i made it as a new object is because G.SHADERS is a table of userdatas, so you can't add a function for getting send vars from one of those, but SMODS.Shaders doesn't contain vanilla shaders, so the best solution i thought of was to separate the initialization of a shader and adding it to the screen (though that still can be done via the `path` property)
i'm open to changing it if it's needed, but that was just my thought process

cg-223:
~~also restarting the game from the crash screen does not work with this pr (it does work on main branch)~~

cg-223 (approved):

PR #1178, opened 2026-01-24
SMODS.should_keep_on_use for finer control over keep_on_use

borb43:
Adds `SMODS.should_keep_on_use(card)`, which now handles calling a center's `keep_on_use`. This can be hooked to more easily allow external effects to keep used consumables, whereas previously all mods would have to patch the same piece of code to achieve this.

## Additional Info:


Aurelius7309:
The more practical approach would probably to either integrate keep_on_use into the using consumable context or make a new context for it. Hookables for effects like this aren't all that desirable

PR #1184, opened 2026-01-30
Prospective #1168 Fix

### LINKED ISSUE #1168: Debuffed cards not counted in `context.individual and context.cardarea == G.hand`
See title. 

Inconsistent with vanilla functionality. Easiest to see with Baron not displaying the "Debuffed" message over debuffed kings in hand.

Gloompashcy:
Fixes #1168 by altering the `context.other_card:can_calculate` check in `eval_card` from exiting early to setting a context flag to true

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -44,7 +44,7 @@ function eval_card(card, context)
         return {}, {}
     end
 	local vanilla_debuff_handling
-    for _,name in ipairs({'Shoot the Moon', 'Baron', 'Reserved Parking', 'Raised Fist'}) do if card.ability.name == name then vanilla_debuff_handling = true; break end end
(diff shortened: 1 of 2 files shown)


Eremel:
I think changing this will unintentionally make any modded joker that uses this context now trigger on debuffed cards. I really don't want to have to make devs check for debuffed cards themselves, so I'm thinking that the way to fix this issue is to hardcode a Baron workaround and let every other calculation continue as it always has. I'm open to other thoughts though.

Gloompashcy:
I was unsure about following through with hard-coding a work-around since there's 4 base game jokers affected by this issue (Shoot the Moon, Baron, Reserved Parking and Raised Fist). But since it's worked like this for so long, it did seem like the only surefire way to avoid messing with modded jokers was to hard-code it. Honestly, this was mostly me attempting to find a fix without hard-coding, but it does seem to be the only option here.

PR #1186, opened 2026-02-01
Preflight V1

WilsontheWolf:
This adds a new mechanism I've dubbed "preflight". It's also bundled up with lovely 0.9.0 changes (0.9.0 is required), as it depends on a few of the new features. In practice this PR implements:

- Preflight loading
  - Mods are loaded (not run) before the game starts, and if there are conflicts of a lovely module, it will disable it and restart the game.
  - (note theres probably some jank around multiple mods in the same directory and this system. I decided it's not worth my time to fix this one somewhat rare case).
- ZIP loading
  - Replaces #985 
  - Works fully, including SMODS being a zip, enabling/disabling zips
  - This requires a modifed NFS. I've renamed our nativefs module to a module called `SMODS.nativefs` and can be accessed at the global `SMODS.NFS`. This is to prevent a mod vendoring NFS and overriding our global breaking zip loading. It should probably be mentioned in the release notes to use SMODS.NFS. (The global NFS and nativefs modules still exist for backward compatibility)
- Restarting uses love.event.quit("restart")
  - This was enabled by changes in lovely 0.8.0, but we waited for 0.9.0 because we could reload patches then reload like this.
  - This should fix some jank with the old system such as double consoles, detaching from the process in terminals, better mobile support.
  - If mods make threads, this might break. I'm not aware of mods that do make threads, and preflight v2 will probably have some sort of thread helper so mods can be torn down correctly.
- A WIP ui system.
  - This is designed to be used for the preflight loading screen and a new crash handler, but due to wanting to push the lovely changes asap, the code using this was commented out. V2 will be using this, but I didn't want to remove the code for now. For review, preflight/sharedUI is not used

Couple things I'm going to do right after I make this (didn't want to delay for this)
- [x] Rebase on latest changes (probably will get squashed)
- [x] Proper error message for older lovelys
- [x] Clean up some files

## Additional Info:

### DIFF
--- a/src/crash_handler.lua
+++ b/src/crash_handler.lua
@@ -492,6 +492,18 @@ end
 -- Note: The below code is not from the original StackTracePlus.lua
 local stackTraceAlreadyInjected = false
 
+local function doRestart()
+    if SMODS and SMODS.restart_game then
+        SMODS.restart_game()
+    else
+        local test, msg = pcall(function()
+            require"lovely".reload_patches()
+        end)
+        if not test then sendErrorMessage("Failed to reload patches... " .. tostring(msg), "StackTrace") end
+        love.event.quit("restart")
+    end
+end
+
 function getDebugInfoForCrash()
     local version = VERSION
     if not version or type(version) ~= "string" then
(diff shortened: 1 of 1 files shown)


WilsontheWolf:
Fuck it we ball

PR #1193, opened 2026-02-04
Hidden consumables polling functions

nh6574:
Adds `SMODS.poll_soul(args)`, function to poll hidden consumables that returns the key, can return nil if none is selected. It takes the following arguments:
* `set: string` set to check if the hidden consumable should replace.
* `key?: string` random seed key
* `mod?: number` multiplies the hidden consumable rates
* `guaranteed?: boolean` always returns a key (if there's a valid consumable)
* `allow_duplicates?: boolean` act as if Showman is present
* `ignore_vanilla?: boolean` ignores Soul and Black Hole

Also `SMODS.get_soul_rate(consumable, set)` that gets the rate of a given consumable (0 if none, nil if invalid). Which also uses `G.GAME.soul_mod` to modify the rate globally.

Consumables now additionally take a table of sets for `soul_set` and a function for `soul_rate`.

As a bonus, now modded hidden consumables can appear even if The Soul is banned (and respect their own bans).

There's some weird jank around Soul and Black Hole that I decided not to touch much to not mess with vanilla functionality so they don't follow the same logic as the rest (hence `ignore_vanilla`). I'm also bad at math at 1 AM so I might have messed with some of the rate calculations in a way I don't realize.

## Additional Info:


nh6574:
I need to test this a bit more and add support for the new weights feature. I will work on it once I have the time (or energy)

I would also appreciate feedback : )

PR #1197, opened 2026-02-06
`context.modify_final_cashout`

the-Astra:
Adds `context.modify_final_cashout` which is a context that triggers after `calc_dollar_bonus` calls that allow for manipulation of the final cashout reward. Has `context.dollars` to represent the final cashout value at time of trigger. Returning `cashout_mod = value` in a return will modify the cashout value by the returned value and automatically add a row to the cashout report. `context.dollars` is updated accordingly after a trigger

## Additional Info:


Aurelius7309:
I don't love to be adding more single-use functions when this is a perfect use case for a self-modifying context

the-Astra:
> I don't love to be adding more single-use functions when this is a perfect use case for a self-modifying context

The original intent was for this to be a context, but I wasn't sure how to do the automatic cashout lines through a method like that

PR #1207, opened 2026-02-11
SMODS.get_table_keys

InvalidOS:
Turns the local `get_keys` function from `SMODS.upgrade_poker_hands` into a separate utility function, `SMODS.get_table_keys`, since it can be useful in some circumstances for modders to have access to. It takes a table as input and returns a table containing all of that table's keys as values.

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -3227,6 +3227,15 @@ function SMODS.blind_modifies_draw(key)
     if SMODS.Blinds.modifies_draw[key] then return true end
 end
 
+-- Gets all keys from a table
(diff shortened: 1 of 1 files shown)


Aurelius7309:
The local `get_keys` function has no reason to exist in the first case. In both of the cases where it's used, other tables containing the constructed information already exist in the global state. Neither do I see any reason for smods to have this kind of util function, given you probably need to revise your data structures if you need it, and it's neither complicated nor tedious to write. Closing in favor of using pre-existing data.

PR #1213, opened 2026-02-16
Add `.lovelyignore` to `.gitignore`

nh6574:
pls john smods

## Additional Info:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,3 +200,6 @@ core/EditionAPI.lua
 # LSP/Editor settings
 .vscode
 .luarc.json
+
+# Balatro specific
+.lovelyignore


DigitalDetective47:
smods now uses lovely's blacklist feature instead of `.lovelyignore` files, so this shouldn't be necessary

nh6574:
SMODS doesn't disable itself, so that wouldn't matter either way. This is for adding the file manually. 

Of course, this could also be solved by using the blacklist but I feel like rejecting this simple PR for that feels nitpicky

Aurelius7309:
No harm to be done in this one. Even if one does use the blacklist, it would never make sense for someone to commit a potentially present .lovelyignore file. Merging

PR #1216, opened 2026-02-18
fix and optimize sounds and music (on macos?)

nnmrts:
On MacOS, with enough mods enabled which load sounds and/or music, the game goes completely silent. I think I have tracked down the root issue for this: OpenAL has a maximum playing instances limit (see https://github.com/MonoGame/MonoGame/issues/4403). And if I understood the code correctly, vanilla loads and plays all songs simultaneously and just changes the volume when switching from one song to another. This is fine for the limited songs in vanilla, but seemingly exhausts that OpenAL limit with mods. I couldn't find much more info on this, but this fix, to only load vanilla songs and the one `desired_track` actually needed, at least worked on my machine. 😝 This may cause either "unexpected" additional calls of `RESTART_MUSIC` or even delays, when a mod tries to play a song mid-game. But I guess that's better than silence.

The `should_stream` thing was my first attempt at fixing this when I first thought this was a memory issue...it still seems useful, because if I understood everything correctly, it should avoid sounds being fully decoded into memory in runtime. Feel free to remove it from this PR, if this is too out-of-scope.

I tested neither of those on a Windows PC though, so I wouldn't merge this before checking that.

## Additional Info:

### DIFF
--- a/lovely/sound.toml
+++ b/lovely/sound.toml
@@ -98,7 +98,7 @@ elseif request.type == 'sound_source' then
     SMODS_Sounds[request.sound_code] = {
         sound_code = request.sound_code,
         data = request.data,
-        sound = sound,
+        should_stream = request.should_stream,
         per = request.per,
         vol = request.vol,
     }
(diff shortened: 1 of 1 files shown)


nnmrts:
This PR causes https://github.com/SiIvaGunner/qualatro to crash on start because of this: https://github.com/SiIvaGunner/qualatro/blob/3661d6472b83008a6d8ea3f7ff0e06aa122e4dff/lovely/Music.toml#L86-L252

nnmrts:
Just pushed a commit to fix the qualatro crash. The fix is ugly, I know...

PR #1244, opened 2026-03-06
`calc_dollar_bonus` improvements

nh6574:
Adds `calc_dollar_bonus` for "individual" objects (by default decks, blinds, challenges, stakes and mods). 

For this I added the set and key of the object to its table, which shouldn't break any functionality for any mod adding objects since it is only used if they add a `calc_dollar_bonus` function to it. Although it could be useful for identifying the objects in the future.

It additionally adds a second `table` return to the functions that allows some styling manipulation for the name. 
  * `text`: sets the text directly
  * `key`, `set`: changes the key and set for the localization (ignored if `text` is set)
  * `text_colour`, `scale` 

## Additional Info:


NeatsTopFoo:
Would you be willing to add a ``no_eval_row`` condition to the ``add_round_eval_row()`` calls that parses a ``no_eval_row`` arg in one of the return tables from ``calc_dollar_bonus`` so that we can make use of the ability to call ``add_round_eval_row()`` in ``context.round_eval``?

Like
```lua
if not ret.no_eval_row then
    add_round_eval_row({dollars = ret, bonus = true, name='joker'..i, pitch = pitch, card = _card, loc_opts = ret_opts})
end
``` 

This would allow for us to do things like writing our own custom round eval rows via ``name = 'custom'``

nh6574:
Makes sense, will do

PR #1055, opened 2025-11-01
(VERY WIP) The Modular Ante Path and its consequences (on my sanity)

AllUniversal:
This is a heavily work-in-progress PR to implement a `AntePath`, requiring a rework of the Blind Select screen (and game states in general) + other subfeatures. I'll write a detailed breakdown when I'm done, feel free to ask if you need something clarified. :)

EDIT 2026.4.5: `SMODS.GameState` and related has been moved into its own PR #1307.

## Additional Info:

- [ ] I didn't modify api's or I've made a PR to the [wiki repo](https://github.com/Steamodded/wiki).
- [ ] I didn't modify api's or I've updated lsp definitions.
- [ ] I didn't make new lovely files or all new lovely files have appropriate priority.

PR #1251, opened 2026-03-08
Blind Slots API

icyethics:
Added pooling, reroll and customization functionality to blind slots.

Blind Slots:
Separate layer from blind, determining a bunch of customizable qualities for each slot. Extends any functionality only associated with the boss blind slot to the big and small blind slot, such as having boss blinds, being able to be rerolled, and ending the ante. Also added utility functions to make use of this new separation between blind slot and blind object, such as when checking for a blind's boss status, or whether it should transition the ante. 

Blind pools:
Added simple blind pools. All custom blinds are automatically marked as Boss blinds unless instructed otherwise. All vanilla blinds are marked as boss blinds, except for the small and big blind, which each occupy their own pool. Blinds can have multiple pools, and blinds can be added or removed from pools during a run.

Changes to Skipping:
Skipping can now be disabled and/or hidden. This can be done generally through effects, or defined for each blind slot. Disabling skips makes it so that the UI still appears but is greyed out and cannot be clicked. Hiding it removes the skip tag UI from the blind. The Boss blind slot is exempted, and cannot have a skip tag be enabled.

Changes to Rerolling:
Blind Slots can be individually rerolled. Enabling a slot's reroll functionality will reroll those slots when the existing reroll function is used, like the boss reroll voucher's or the reroll tag's. The new blind slot specific reroll function will always be able to reroll a blind.

Blind Slot Altering & Disabling:
Blind Slots are now set up to allow for their own blind requirement modification, which will be applied separately from the blind object's. Modifications can be a multiplier or a hard value setting, ignoring source value. Slots can also be set up to disallow disabling of the blind object, or disallowing altering the blind requirement. Blind requirement altering can now be done with SMODS.alter_blind_requirement()

Showdown Antes:
The Showdown ante is now configurable. Additional ante numbers can be set to allow for showdown blinds too, and the ability for a blind slot to be able to spawn showdowns can be set separately for each Blind Slot.

## Additional Info:

- [x ] I didn't modify api's or I've updated lsp definitions.
- [x ] I didn't make new lovely files or all new lovely files have appropriate priority.

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -839,7 +839,7 @@ payload = '''
 SMODS.saved = false
 G.GAME.saved_text = nil
 SMODS.last_hand = SMODS.last_hand or {scoring_hand = {}, full_hand = {}}
-SMODS.calculate_context({end_of_round = true, game_over = game_over, beat_boss = G.GAME.blind.boss, scoring_hand = SMODS.last_hand.scoring_hand, scoring_name = SMODS.last_hand.scoring_name, full_hand = SMODS.last_hand.full_hand })
+SMODS.calculate_context({end_of_round = true, game_over = game_over, beat_boss = SMODS.blind_is_boss(), beat_boss_exception = SMODS.blind_is_boss(true), ends_ante = SMODS.blind_ends_ante(), scoring_hand = SMODS.last_hand.scoring_hand, scoring_name = SMODS.last_hand.scoring_name, full_hand = SMODS.last_hand.full_hand })
 if SMODS.saved then game_over = false end
 -- TARGET: main end_of_round evaluation
 '''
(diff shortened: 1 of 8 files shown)


AllUniversal:
Very cool! I planned a different approach in #1055, replacing the structure of an Ante entirely, but that PR isn't being completed soon I fear. :') (Unless someone wants to add a pannable, map-like UI element)

icyethics:
I was looking into how to make this compatible with the new weight system. Some features this adds have been added by other PRs, so the blind requirement manip functionality can be thrown out. 

The way I'd set up pools has been slightly integrated, I think? The latest smods does at least use a similar set up. Because my PR was written before attributes were a thing, it doesn't use those, but I think the best way forward would be to use blind_pools as the table of keys, and use the attribute system to poll them. 

@AllUniversal It's not part of the PR besides some exposed ends for someone else to make use of, but I'm sure there might be a more effective way to handle the 'transitions_into' argument that I set up, so that a blind tree system could properly use them.

PR #1252, opened 2026-03-08
fix scoring calculation operator text not being set when re-entering a run

InvalidOS:
noticed this when i was fucking around with scoring calculations earlier. i had set the scoring calc to exponent, and when re-entering the run the ui was the default multiply ui, but the current calc was still exponent

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -2776,6 +2776,7 @@ function Game:start_run(args)
         trigger = 'immediate',
         func = function()
             SMODS.refresh_score_UI_list()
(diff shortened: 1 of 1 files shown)


nh6574:
I did a different fix for this in 6e5beb6c5c470b7c8e4ab26b9b6c0769388c8e44 which doesn't involve an event so it does it immediately. Let me know if that solves the issue.

edit: oops the commit had a logic error. fixed :p

InvalidOS:
yep this works, ty

PR #1255, opened 2026-03-11
Adjust Chaos the Clown

NeatsTopFoo:
Makes Chaos the Clown actually use its extra value instead of just being a flat 1; This improves flexibility if a mod wanted to say, scale its reroll count using ``scale_card`` as part of a mechanic effect or some other interaction.

## Additional Info:


nh6574:
I think the idea makes sense but I feel like this could lead to a problem where it removes less or more than what it adds.

PR #1264, opened 2026-03-18
fix: Android ScreenShader — canvas bug + perf improvements

dddfault:
## Additional Info:

### DIFF
--- a/version.lua
+++ b/version.lua
@@ -1,1 +1,1 @@
-return "1.0.0~BETA-1527a-STEAMODDED"
+return "1.0.0~BETA-1527b-STEAMODDED"


dddfault:
Fix rendering bug and major performance issues when using `SMODS.ScreenShader` on Android. PC behavior unchanged.

## Fixes

### 1. Canvas size (1/4 screen bug)
On Android, `fromPixels()` + `CANV_SCALE` caused double scaling → canvas rendered at 1/4 size.

**Fix:** use `getPixelWidth/Height()` on Android.

---

### 2. Performance (extra blit)
Ping-pong pipeline forced multiple fullscreen draws per frame → heavy on mobile GPU.

**Fix:** Android now uses direct draw (no intermediate canvases, 1 draw call).

---

### 3. CRT overhead
CRT uniforms + calculations still ran every frame even when not applied.

**Fix:**
- Disable CRT when custom ScreenShader is present (`obj_buffer > 1`)
- Skip `send_vars` on Android when inactive
- Strip inline CRT processing from `game.lua` via regex patch

---

## Result

| Platform | Custom Shader | CRT | Pipeline | Overhead |
|---|---|---|---|---|
| PC | Any | Active | Ping-pong | Same |
| Android | No | Active | Direct | Low |
| Android | Yes | Disabled | Direct | Low |

---

## Impact
- Fixes 1/4 screen rendering bug on Android
- Restores FPS (~20 → ~30+)
- Removes redundant per-frame CRT cost
- No regression on PC

real-niacat:
did a quick test, the change from ping-pong to single-draw appears to not render any actual ScreenShaders other than the CRT if is_android is forced on, this change *is* beneficial if only one shader is active, but otherwise it does miss the others, even if the CRT is disabled

intended behavior (preserved on PC still) - screen is darkened from the CRT and the additional ScreenShader is properly rendered
image

real behavior (simply doing `is_android = true` to force single-draw) - screen is properly darkened by CRT if enabled, but other shader has no effect, regardless of CRT state
image


let me know if i missed something or tested improperly, but from what i see this doesn't recreate ping-pong fully, otherwise, the canvas scale changes seem good

dddfault:
This change is intended to mitigate the issue by ensuring that only a **single custom ScreenShader is active**, effectively overriding the game's default CRT shader.

I chose this approach because there is little benefit in using multiple layers or ping-pong shaders, especially on Android devices particularly with weaker GPU performance.

By overriding the game's default CRT shader, the performance overhead is reduced further, resulting in higher and more stable FPS.

Aurelius7309:
It would be better to not fully disable the option to have more than one screen shader active. I suggest the pipeline be adjusted depending on if additional screen shaders are active and/or a user-facing setting to limit screen shaders on android devices with a weaker GPU

real-niacat:
a large priority behind screenshader initially was being able to stack them dynamically, as to avoid prioritizing one mod, since beforehand, you had to completely replace the CRT which was very mod-unfriendly

replacing it with single-draw if only one is active makes perfect sense, but if there are multiple i believe it should preserve intended behavior rather than performance (though, a config option to prefer 1 single-draw over ping-pong would be fine, but this does entirely defeat the purpose)
additionally, it doesn't seem to work on actually drawing custom shaders if the CRT is disabled

dddfault:
Good point. I'll update the pipeline soon, in the mean time I still have no time to do that. 

It'll ignore the single-draw pipeline shader mode if there is more than 1 ScreenShader active.

dddfault:
@real-niacat
I just did a quick test on Windows by applying same condition on the changes. Everything is working as intended.

The issue was on `screenshader_rendering.toml` patch, it directly use single-draw pass on Android instead of using ping-pong if there is more than 1 ScreenShader exist.

and what do you mean by "screen is properly darkened by CRT if enabled" ?

real-niacat:
make sure to account for the fact that screenshaders can exist without being drawn (via ScreenShader.should_apply)
and yeah i was a bit vague, my bad, i just meant that the CRT is properly drawn (and that the CRT's effect just tends to increase contrast and darken the screen)

dddfault:
@real-niacat 

Ah I see, that color correction on default CRT shader. Most of people (especially player with weaker Android devices) prefers default CRT shader disabled entirely. That's why I applied this approach to SMOD ScreenShaders. Since I can apply my own simplest way to do color correction.

This should be a simple changes on `screenshader_rendering` patch. Soon.

PR #1273, opened 2026-03-24
feat: add conditionals for cycling dynatext content

Foo54:
Adds the ability to limit when certain content can appear in a cycling DynaText

### Example
```lua
DynaText({
  string = {
    "Did you know ... ",
    {string = "You're rich!", condition = function() return G.GAME.dollars > 50 end},
    {string = "You're poor!", condition = function() return G.GAME.dollars <= 50 end}
  }
  -- other code
})
```

## Additional Info:

### DIFF
--- a/lovely/dynatext_conditionals.toml
+++ b/lovely/dynatext_conditionals.toml
@@ -44,3 +44,21 @@ end
 '''
 match_indent = true
 
+# Check condition for the initial string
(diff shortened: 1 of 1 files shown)


nh6574:
I added some code so the first string doesn't show up if the condition isn't met, but otherwise looks good!

PR #1276, opened 2026-03-27
fix: `SMOD.ScreenShader` bugfix and performance improvement for Android build

dddfault:
Improve Android performance and rendering behavior for `SMOD.ScreenShader`.

- Fixed render size issues affecting Android builds
- Optimized pipeline:
  - Use single-draw pipeline when only one ScreenShader is active
  - Fallback to ping-pong rendering when multiple shaders are active (matching original behavior)
- Reduced overhead in CRT parameters

These changes improve performance and ensure consistent behavior across Android devices.

## Additional Info:


Polprzewodnikowy:
Same issue exist on macOS and every device that has `window.highdpi` option enabled. Canvas scale should be read from `love.window.getDPIScale()` function. No need to make system specific hacks.

https://love2d.org/wiki/love.window.getDPIScale
https://love2d.org/wiki/Config_Files#:~:text=monitors%20are%20available.-,window.highdpi,-Available%20since%20L%C3%96VE

Or even better remove `love.window.fromPixels()` since `love.graphics.getWidth()/getHeight()` already return correct value without multiplying by DPI scale.

Aurelius7309:
ScreenShader performance changes need further review. The bugfix works, applying it on main separately from this PR

dddfault:
for the CRT performance pass need review from other user especially on lower end mobile devices. because on my end it's confirmed. CRT being applied twice. first from the game itself, and via SMOD.ScreenShader function.

dddfault:
Need suggestions for any user that played this game on mobile devices. Do you really use game CRT shader? if so, then the current implementation of single-pass kinda useless since CRT shader now called via SMOD.ScreenShader, so the next custom shader will call ping-pong rendering and that's gonna tank performance on lower end mobile devices.

Also I changed to use G.CANVAS:getDimensions(), this will fix ping-pong rendering behavior on high-dpi, no more skewing and weird canvas rendering artifact. Tested on Windows and Android.

Abdallahpr1:
I play on my phone and the first thing that i did when opening Balatro was to go to settings and disable the CRT effect
 
Also how can i download the files that you modified ?

Aurelius7309:
@Abdallahpr1 you can download from here: https://github.com/dddfault/smods/archive/refs/heads/main.zip

Abdallahpr1:
@Aurelius7309 thanks
 
Tried it, it gives the same bug the unmodified smods version does
@dddfault 
Screenshot_2026-04-08-11-22-29-189_systems shorty lmm balatro

dddfault:
Hmm.. prolly the `getDimension()` thing but in my end it's all good. try to disable any mods and just launch only with SMOD enabled.

Abdallahpr1:
Same thing

dddfault:
Your issue is a different one than canvas rendersize. the problem might be on the internal game shader itself. CRT might be the causes but I cannot confirm. Because everything i tried to fix was SMOD.ScreenShader() features and your issue wasn't exist until version 106x

Abdallahpr1:
It's likely because of my phone's soc, poco x7 pro
On youtube someone said that he didn't have any issue when putting smods after version 1016c
While it's not a crt bug i think it's definitely a shader bug

dddfault:
oh yeah, I remember it. I have encountered same issue on UNISOC phones. definitely it's game shader because i ran my own custom shader with SMOD.ScreenShader() function and haven't encountered that issue so far.

Abdallahpr1:
If you found a fix pls ping me so i can try, if not I'll just stick to version 1016c

PR #1277, opened 2026-03-28
DynaText shaders, more robust UI element shaders

InvalidOS:
Does the following:
* Adds the ability for UI shaders to be applied to text, and tell whether they're being applied to the main text or the text shadow.
* Adds a `shader` config option to DynaText
* Adds the ability for DynaTextEffect to apply shaders
* Optimizes UI Element shaders

Notable limitations:
* Text shaders don't work with underlined text
* Shaders are applied to an entire UI element indiscriminately, with the exception of text.

These issues are likely to be fixed later, but given the presence of UI shaders on the main branch, having a working implementation is probably a better idea.

## Additional Info:


Aurelius7309 (approved):
lgtm (i totally reviewed all the code)

PR #1284, opened 2026-03-29
feat: Add `weights` to every object and `SMODS.Attribute`

Eremel:
This branch adds two new features:
- a functional `weight` property on each object that is respected when polling items
- `SMODS.Attribute` which can be used to easily group and poll sets of objects *(primarily used for Jokers but could be used for other card types too)*

## Weight system
This system changes how objects are polled in every (?) area of the game - it does not respect vanilla seeding so is enabled by an optional feature, `object_weights`.

Adding `weight = X` to an object definition allows you to customise the weight.

Using the new function `SMODS.poll_object`, you can get a key that respect item weights. This function can combine different pools and rarities too. It accepts a table to control the function.

```lua
type = String -- specify a single type (or attribute) to poll
types = Table{String} -- specify a table of types (or attributes) to poll
attributes = Table{String} -- specify a table of types (or attributes) to poll
pool = Table{String} -- specify a table of object keys to use as the pool instead of having the function generate one
filter = function(pool) -- specify a function to filter the pool once it has been generated
intersect = Boolean -- decide whether multiple pools are intersect instead of union

guaranteed = Boolean -- guarantee a result from the function
chance = number -- % of success as a decimal value 0-1
seed = String -- specify a seed to use for the poll chance (defaults are available)
type_key = String -- a second poll seed if repolling is used
append = String -- appended onto the poll seed
no_ngeative = Boolean -- swaps negative edition to polychrome for vanilla functionality

print = Boolean -- turns on debug prints
```

*There is some initial work on making custom Blinds available for other slots here. I need to look at how #1251 can fit in with this.*

## `SMODS.Attribute`
This is intended to be used alongside `SMODS.poll_object` to allow for sets of cards to be polled more easily. Requires a String `key` and accepts a Table{String} `keys` of object keys.
Also accepts a Table{String} `alias` to allow for alternate names.

Other objects can add `attributes = {'attribute_key', 'attribute_key_2'}` to their definitions to be added into each pool. To add many objects at once, `SMODS.add_attribute(attribute_key, object_keys)` exists, where `attribute_key` is the key of the attribute, and `object_keys` is a table of keys that would be used in the definition.

TODO:
- any clean up *(there are some extra debug things hanging around still)*
- verify weight polls are consistent per seeds (not vanilla)
- verify weight polls aren't used when feature is disabled
- verify attribute lists are correct

## Additional Info:


Eremel:
TODO: add a way for attribute pools to respect rarity

Aurelius7309 (approved):

PR #1253, opened 2026-03-09
Implement SMODS.GUI.scrollbar and SMODS.GUI.dropdown_select

ThunderEdge73:
# `SMODS.GUI.scrollbar` and `SMODS.GUI.dropdown_select`
This PR adds `SMODS.GUI.scrollbar(args)`, a function that returns a UI node that can be scrolled. It can optionally be linked to an `SMODS.UIScrollBox` in `args.scroll_collision_obj` to automatically handle scrolling that object. Otherwise, it will update `args.ref_table[args.ref_value]` based on the scrollbar's progress. Finer control over the specific UI node type returned is possible via `args.ui_type`. This also adds `SMODS.GUI.dropdown_select`, a cleaner way to select an option from a list.

## Why not use `create_slider` instead of `SMODS.GUI.scrollbar`?

`create_slider` creates a slider that also has a numeric display next to it by default, whereas `SMODS.GUI.scrollbar` does not. Furthermore, `create_slider` does not support mouse scrollwheel inputs, but `SMODS.GUI.scrollbar` does. `SMODS.GUI.scrollbar` also has a knob to indicate more clearly that it can be dragged and scrolled, much like how it is on websites.

## Examples

For `SMODS.GUI.scrollbar(args)`:

```lua
-- Copied from my personal mod, Multiverse
function Multiverse.credits_tab_definition()
	local rows = {}
	local contributor_text = DynaText({
		string = localize("mul_contributors"),
		colours = { G.C.UI.TEXT_LIGHT },
		shadow = true,
		float = true,
		silent = true,
		spacing = 5,
		scale = 1,
		rotate = true,
		pop_in = 0,
		text_effect = "mul_ui_multiverse_highlight",
	})

	rows[#rows + 1] = {
		n = G.UIT.R,
		config = { align = "cm", padding = 0.1 },
		nodes = {
			{
				n = G.UIT.O,
				config = {
					object = contributor_text,
				},
			},
		},
	}
	for _, entry in ipairs(Multiverse.credits_table) do
		if entry == "MISC_CREDITS" then
			local inspiration_text = DynaText({
				string = localize("mul_inspirations"),
				colours = { G.C.UI.TEXT_LIGHT },
				shadow = true,
				float = true,
				silent = true,
				spacing = 5,
				scale = 1,
				rotate = true,
				pop_in = 0,
				text_effect = "mul_ui_multiverse_highlight",
			})
			rows[#rows + 1] = {
				n = G.UIT.R,
[...]

PR #1287, opened 2026-03-30
Fix scrollbar bug with scrollbox content being smaller than or the same size as scrollbox bounds + Improve Dropdown API

ThunderEdge73:
Exactly what the title says. Fixes a bug that was present in #1253 regarding the scrollbox content rapidly sliding out of view if the scrollbox content was smaller than the size boundaries that the scrollbox defines, or if the scrollbox had no size bounds defined.

Also adds an improvement to the Dropdown API. Now, in the args table, you can define display_choice_func, a function that takes in an option and returns a string that will be displayed instead of the raw value in args.ref_table[args.ref_value].

## Additional Info:

### DIFF
--- a/lovely/better_calc.toml
+++ b/lovely/better_calc.toml
@@ -180,7 +180,7 @@ if context.cardarea == G.hand and context.main_scoring then
 
     local h_score = card:get_bonus_h_score()
     if h_score ~= 0 then
-        ret.playing_card.score = h_score
+        ret.playing_card.h_score = h_score
     end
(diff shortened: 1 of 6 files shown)


ThunderEdge73:
idk why the commit history is so messed up, man i hate git

ThunderEdge73:
I also added an improvement to the Dropdown API. Now, in the args table, you can define display_choice_func, a function that takes in an option and returns a string that will be displayed instead of the raw value in args.ref_table[args.ref_value].

PR #1289, opened 2026-03-30
Pool badges

Evgast:
Badges for pools.
That's how: you add one ::::
image
image

I didn't know where to put that patch so I went with the end of the ui.toml idk

Probably not the most elegant way to implement that? Question mark?

I don't know how to interact with these boxes yeah
## Additional Info:


NeatsTopFoo:
This functionality could be good in tandem with #1284 where you could specify a custom weighted pool and optionally give it some badge

nh6574:
Made some changes to make this into set_badges like Centers have

PR #1299, opened 2026-04-02
Added card `matcher` functions and related.

AllUniversal:
+Title, I moved this over from #838, the latest two comments on that PR explain card `matchers`.

## Additional Info:


AllUniversal:
I now added a new subcondition `count`. It should make it possible to match for cards based on how many cards share a property, like matching a card only if its seal appears `exact`ly once, or `at_least` twice, or `at_most` (=> subflags of `count`) three times, or if a custom `func` (subflag of `count`) returns `true` when called with the total number of the card's property.

EDIT: Oh, and just to mention it here, I also added `Card:get_suits()`.

AllUniversal:
Realized stuff ain't working right, will look at it tomorrow.

AllUniversal:
Fixed (and tested) it, `.count` should now be working!

Aurelius7309:
Ran into some hard crashes while testing it out, so I just went ahead and fixed those ^^
I'm still not close to having fully tested it, but I did notice this not working as expected:
`SMODS.create_card_matcher { enhancement = { any = { m_gold = true, m_steel = true }, count = { at_least = 3 } } }` - in fact I couldn't get it to fire at all. Is this user error on my part or a bug?

AllUniversal:
@Aurelius7309 The following code works for me; 
image
Specifically, this matcher would match a card if; 
1. It is either a Gold or Steel card
2. Its enhancement (*no matter what*) must appear at least three times.
-> Thus, in a hand of three or more Gold cards and less than three Steel cards, it matches every Gold card, but no Steel card. 

If you want it to match three cards that are *either* Gold or Steel *each*, I don't think this is currently supported actually, in which case I'll look into how I'd add it.

Aurelius7309:
Huh, I placed this matcher into the flush Poker Hand Part and it did nothing even with 3 gold + 3 steel cards. Might have done it wrong, I guess I'll experiment some more tomorrow. Also without looking at code I'd have expected the count to relate to the property and not the actual matched card - though either can be useful.

I suppose this means there's also currently no support for linking properties, e.g., two cards of the same rank and suit?

AllUniversal:
Indeed, hmm. I'll see about that too.

AllUniversal:
EDIT 2026.5.28: This is outdated now, the `any_related` subflag has been removed. The below functionality should be achievable by checking the number of cards matched instead.

> If you want it to match three cards that are _either_ Gold or Steel _each_, I don't think this is currently supported actually, in which case I'll look into how I'd add it.

This is now supported through the use of the new `count` subflag `any_related`; It makes it so that if a card matches a `matcher.condition`'s `any` flag, and that `condition` (e.g. `enhancement`) is `count`ed, the `matcher` also adds 1 to all the other values of the `any` flag, thus essentially merging their counts into one.

So the above example of matching `at_least` three cards that are *either* Steel or Gold is now possible with this matcher;
```lua
local matcher = SMODS.create_card_matcher({enhancement = {any = {m_gold = true, m_steel = true}, count = {at_least = 3, any_related = true}}})
```

AllUniversal:
@Aurelius7309 Aight so I added a new `count` subflag `overlap`; It too has the `all`, `any` or `none` (new) subflags, but for conditions, not keys. This is a `matcher` for matching two or more cards, as long as they have a Red seal and share enhancements;
```lua
SMODS.create_card_matcher({seal = {any = {Red = true}, count = {at_least = 2, overlap = {any = {enhancement = true}}}}})
```
One for matching exactly two cards with a Red seal and shared enhancement;
```lua
SMODS.create_card_matcher({seal = {any = {Red = true}, count = {exact = 2, overlap = {any = {enhancement = true}}}}})
```
(This one would *not* match two cards with Red seals if there's a third one, *unless that one has a different enhancement*)

AllUniversal:
Was just inspired to try implementing **Holy Fool**'s Faust Joker's **Devil's Hand** pokerhand using card matchers, this is the code (replacing Flush for testing):
```lua
SMODS.PokerHandPart {
    key = '_flush',
    func = function(hand) 
        local matcher = SMODS.create_card_matcher({suit = {count = {exact = 1}}, rank = {count = {exact = 1}}})
        local m_t_c, c_t_m = SMODS.match_cards(hand, {matcher})
        return SMODS.get_hand_from_matching(m_t_c, c_t_m, {matcher_max_cards = {[matcher] = 4}})
    end -- get_flush(hand) end,
}
```
Making these four cards count as a Devil's Hand / Flush:
image

AllUniversal:
I have now added subflags to `count.overlap[condition]`; 
```lua
"any" = boolean             -> any [condition]s overlap
"all" = boolean             -> all [condition]s of the primary card checked overlap with the card checked against
"all_either" = boolean      -> all [condition]s of the card with fewer [condition]s overlap with the other card 
"none" = boolean            -> none of the [condition]s overlap
"exact" = integer           -> [exact] number of cards overlap
"at_least" = integer        -> [at_least] number of cards overlap
"at_most" = integer         -> [at_most] number of cards overlap
"func" = functions          -> see `count.func`
```
These allow checking overlap more precisely, e.g.; 
```lua
{suit = {count = {at_least = 3, overlap = {any = {enhancement = {none = true}, seal = {none = true}}}}}}
```
^ This is a matcher to match *at least 3* cards which share a *suit*, as long as they either share *no* enhancements or *no* seals.

```lua
{suit = {count = {at_least = 3, overlap = {any = {enhancement = {none = true}, seal = {any = true, at_least = 3}}}}}}
```
^ This is a matcher to match *at least 3* cards which share a *suit*, as long as they either share *no* enhancements or a *seal* with *2* other cards (with which they share a suit).

PR #1300, opened 2026-04-03
Add assert to shader loading for missing files

lord-ruby:
## Additional Info:


Aurelius7309:
`attempt to call global 'NFSread' (a nil value)`

PR #1305, opened 2026-04-04
Ability to disable calls to all joker's calc_scaling functions in smods.scale_card

Foo54:
Moderately useful, but lightweight idea I had off of a question in #modding-dev

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -3073,19 +3073,21 @@ function SMODS.scale_card(card, args)
 	        end
 	    end
 	end
-    if card.edition then
-        local edition = G.P_CENTERS[card.edition.key]
-        if edition.calc_scaling and type(edition.calc_scaling) == 'function' then
-            local ret = edition:calc_scaling(card, card, initial, scalar_value, args)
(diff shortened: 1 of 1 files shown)


nh6574:
What's a use case for this? I read the discussion in modding-dev that it was to prevent a loop with scale_card inside of calc_scaling but I don't see how this is useful over a check that the card scaling is not itself or not using scale_card in the first place. Doesn't this prevent the point of the function?

Also, I think it makes more sense for the flag to be something like `block_scaling = true` instead. Adding a flag for this that by default is true is confusing and it's better to mirror `block_overrides`.

Eremel:
I am also struggling to see what a use case for something like this would be. If you can give a concrete example of an effect that would need it, we can take a look at adding it, but as it stands it seems very counterintuitive to what the function is used for.

Foo54:
I made this around midnight when I was really tired, and now that more awake I see no point in ;-;
mb

NeatsTopFoo:
🥀

PR #1320, opened 2026-04-08
fix: resolve nil-indexing crashes in deck_preview UI

spadesmaster:
Adds three critical safety checks to G.UIDEF.deck_preview in src/overrides.lua to prevent "attempt to index a nil value" crashes during modded runs.

Changes:
   1. Logic Fix (Line ~736): Added checks for v.base.suit and its initialization in the SUITS table before indexing. This prevents crashes when encountering non-standard card data.
   2. UI Text Fix (Line ~798): Added a nil-safe check when displaying card counts per rank (SUITS[suit][rank]). This ensures the UI doesn't crash if a specific suit/rank combination isn't initialized.
   3. Palette Fix (Line ~810): Added a nil-check for deckskin before accessing palette_map. This prevents a crash if a suit is mapped to a custom deck skin that is missing or uninitialized.

These fixes significantly improve the stability of the Deck Preview UI in environments where mods (like Cryptid) introduce exotic card properties.
  
## Additional Info:

### DIFF
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -806,7 +806,7 @@ function G.UIDEF.deck_preview(args)
 
 	for k, v in ipairs(suit_map) do
 		if not hidden_suits[v] then
-			local deckskin = SMODS.DeckSkins[G.SETTINGS.CUSTOM_DECK.Collabs[v]]
+			local deckskin = SMODS.DeckSkins[G.SETTINGS.CUSTOM_DECK.Collabs[v]] or {}
 			local palette = (deckskin or {}).palette_map and (deckskin.palette_map[G.SETTINGS.colour_palettes[v] or ""] or ((deckskin or {}).palettes or {})[1]) or {}
 			local t_s
 			if palette and palette.suit_icon and palette.suit_icon.atlas then
--- a/src/overrides.lua
+++ b/src/overrides.lua
@@ -808,6 +808,7 @@ function G.UIDEF.deck_preview(args)
 		if not hidden_suits[v] then
 			local deckskin = SMODS.DeckSkins[G.SETTINGS.CUSTOM_DECK.Collabs[v]]
 			local palette = deckskin and (deckskin.palette_map and deckskin.palette_map[G.SETTINGS.colour_palettes[v] or ""] or (deckskin.palettes or {})[1]) or {}
+			local t_s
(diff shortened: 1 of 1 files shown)


spadesmaster:
Verified these fixes locally in a modded run with Cryptid and high-complexity decks.

Aurelius7309:
`table[nil]` returns `nil`, so checking `v.base.suit` is unnecessary. style: `(a or {}).b` is preferred to `a and a.b` as a nil check.

Looks good otherwise

spadesmaster:
Updated the code to use the preferred (a or {}).b style and removed redundant suit checks as suggested. Thanks for the feedback!

src/overrides.lua
```diff
@@ -807,7 +807,7 @@ function G.UIDEF.deck_preview(args)
 	for k, v in ipairs(suit_map) do
 		if not hidden_suits[v] then
 			local deckskin = SMODS.DeckSkins[G.SETTINGS.CUSTOM_DECK.Collabs[v]]
```
Aurelius7309:
```suggestion
			local deckskin = SMODS.DeckSkins[G.SETTINGS.CUSTOM_DECK.Collabs[v]] or {}
```

src/overrides.lua
```diff
@@ -807,7 +807,7 @@ function G.UIDEF.deck_preview(args)
...
-			local palette = deckskin.palette_map and deckskin.palette_map[G.SETTINGS.colour_palettes[v] or ''] or (deckskin.palettes or {})[1]
+			local palette = (deckskin or {}).palette_map and (deckskin.palette_map[G.SETTINGS.colour_palettes[v] or ""] or ((deckskin or {}).palettes or {})[1]) or {}
```
Aurelius7309:
```suggestion
			local palette = (deckskin.palette_map or {})[G.SETTINGS.colour_palettes[v] or ""] or (deckskin.palettes or {})[1])
```

Aurelius7309 (commented):
The `palette` change doesn't make much sense as-is (it doesn't recognize `palettes` when `palette_map` is absent due to bracketing). Suggested cleanup also moves default empty table into the declaration of `deckskin` as it's used multiple times.

spadesmaster:
OK, accepted the changes you made

PR #1329, opened 2026-04-10
SMODS.scale_card scalar_factor

vitellaryjr:
this makes three changes to `SMODS.scale_card`, which makes it easier to use outside of the most basic cases:
- `args.scalar_factor` can be provided, which will be multiplied with the `scalar_value` in the _operation_ functions specifically (as well as the localized variable). this allows for conveniently recreating the behavior of scaling a card by a multiple of something, which is currently only possible by making a custom operation
- if no `args.scalar_value` is provided, SMODS will automatically make one with a value of 1. the intent of this feature is: combined with the `scalar_factor` argument, you can scale a card by an arbitrary value easily, while still internally using a scalar_value to allow for manipulation from other objects
- `SMODS.scale_card` now returns the final scaled value, and what it was scaled by, to allow jokers themselves to respond / utilize the changed values

## Additional Info:

### DIFF
--- a/lovely/scaling.toml
+++ b/lovely/scaling.toml
@@ -27,9 +27,7 @@ SMODS.scale_card(self, {
     ref_value = "mult",
     scalar_table = sliced_card,
     scalar_value = "sell_cost",
-    operation = function(ref_table, ref_value, initial, scaling)
-        ref_table[ref_value] = initial + 2*scaling
-    end,
+    scalar_factor = 2,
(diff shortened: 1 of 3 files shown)


vitellaryjr:
added another feature to this, having used this function a lot now and finding it would be useful: `SMODS.scale_card` now returns the final scaled value, and what it was scaled by, to allow jokers themselves to respond / utilize the changed values

PR #1332, opened 2026-04-12
`SMODS.PermaBonus`

### LINKED ISSUE #1321: Modularize perma bonuses
*Title, adding custom perma bonuses currently requires patches to multiple targets

Gloompashcy:
Adds the SMODS.PermaBonus class.
This class defines the characteristics of a permanent bonus for cards.

- Required parameters:
  - `key`: The key of the PermaBonus in the `card.ability` table
  - `apply_to`: A string or array of strings telling the game which calculation(s) to apply the bonus to, it can be one of:
```lua
'chips', 'mult', 'x_chips', 'x_mult',
'h_chips', 'h_mult', 'h_x_chips', 'h_x_mult',
'p_dollars', 'h_dollars',
'score', 'x_score', 'h_score', 'h_x_score',
'blind_size', 'x_blind_size', 'h_blind_size', 'h_x_blind_size',
'repetitions'
```
- Optional parameters *(defaults)*:
  - `vars_key`: The key used to store the value in specific_vars when creating the card's description, defaults to the object's key
  - `loc_key`: The key to look for in G.localization.descriptions.Other, defaults to the object's key
  - `signed_value = false`: Defines whether the value should be signed in the description
  - `signed_dollars = false`: Same as above but for money, has priority over signed_value
  - `ui_mod`: Number to add to the displayed value in the card description

- API methods
  - `should_apply(self, card, calculation) -> boolean`
    - Defines the behaviour when Card:get_perma_bonus checks to add the bonus
    - By default it just checks if the `calculation` parameter is equal to the object's `apply_to` value

  - `get_ui_value(self, card) -> number`
    - Defines how the value on the card description is calculated
    - By default it adds the value `card.ability[self.key]` and the ui_mod value (if it exists)
    - Should return `nil` if `card.ability[self.key] == 0`

  - `upgrade(self, card, amount)`
    - Defines how the bonus should be upgraded
    - By default it adds the `amount` parameter to card.ability[self.key]
    - An additional `from` parameter is passed into the function when called through `SMODS.upgrade_perma_bonus`
  
  - `localize(self, value, desc_nodes)`
    - Defines how the descriptions should be localized
    - By default it signs `value` based on the values of `self.signed_value` and `self.signed_dollars`, then calls the vanilla `localize` function like so:
      ```lua
      localize{type = 'other', key = self.loc_key, nodes = desc_nodes, vars = {value}}
      ```

- Changed util functions:
  - `SMODS.localize_perma_bonuses(specific_vars, desc_nodes)`
    - Now iterates through the SMODS.PermaBonus objects and runs each object's `localize` function
    - Iterates through in load order to keep a similar description order to before the changes
    - Does not override vanilla `card.ability.perma_bonus` description
  
- New util functions:
  - `SMODS.set_perma_bonus(card, new_ability)`
    - Runs whenever `set_ability` or `SMODS.quantum_set_ability` create the `new_ability` table
    - Iterates through the SMODS.PermaBonus keys and sets the `new_ability[key]` value to the previous `card.ability`'s value or 0

  - `SMODS.get_perma_bonus_ui_vars(card) -> table`
    - Runs when `generate_UIBox_ability_table` creates the specific_vars table for playing cards
    - Iterates through the SMODS.PermaBonus objects and runs each object's `get_ui_vars` function on the given `card`
    - The returned table is indexed using the objects' vars_key value

  - `SMODS.upgrade_perma_bonus(args)`
    - Upgrades a cards perma bonuses
    - The table `args` can have the following values:
      - `keys` (required): a table of SMODS.PermaBonus keys to upgrade on the card
      - `card` (required): the card that the upgrade will be applied to
      - `amount`: how much to upgrade the bonus by, defaults to 1
      - `from`: the object that triggered the upgrade, if not provided, an empty table is passed into the upgrade function instead
      - `func`: a func to define the upgrade behaviour, overrides all the `upgrade` function from the PermaBonus objects. It takes the following parameters:
        - `card`: args.card
        - `amount`: args.amount
        - `from`: args.from
        - `PermaBonus`: the SMODS.PermaBonus object of the bonus being upgraded

  - `Card:get_perma_bonus(calculation)`
    - Adds together all the perma bonuses for the given `calculation`
    - Is called through each Card:get_value function (e.g. `Card:get_x_mult`) with the corresponding `calculation` value
    - Is called as `Card:get_perma_bonus('repetitions')` in `eval_card`
    - If called without a calculation value, it returns a table of all the perma bonuses on the card, including ones that are 0

Additionally, perma_x_mult and perma_x_chips now correctly say extra in their descriptions.

Not sure if adding a file in game_objects was the correct way to do this, but it seems to be the new standard(?)

Also, this resolves #1321 

## Additional Info:

### DIFF
--- a/src/game_objects/perma_bonuses.lua
+++ b/src/game_objects/perma_bonuses.lua
@@ -10,11 +10,6 @@ SMODS.PermaBonus = SMODS.GameObject:extend{
     prefix_config = { key = false },
     signed_value = false,
     signed_dollars = false,
-    should_apply = function(self, card, calculation)
-        for _,v in ipairs(self.apply_to) do
-            if calculation == v then return true end
-        end
-    end,
     get_ui_value = function(self, card)
         return card.ability[self.key] and card.ability[self.key] ~= 0 and (card.ability[self.key] + (self.ui_mod or 0)) or nil
     end,
@@ -34,7 +29,7 @@ SMODS.PermaBonus = SMODS.GameObject:extend{
     end,
     process_loc_text = function(self)
         SMODS.process_loc_text(G.localization.descriptions.Other, self.loc_key, self.loc_txt)
-    end,
+    end
 }
 
 function SMODS.localize_perma_bonuses(specific_vars, desc_nodes)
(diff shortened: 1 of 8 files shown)


src/game_objects/perma_bonuses.lua
```diff
@@ -0,0 +1,138 @@
+SMODS.Perma_Bonuses = {}
+SMODS.Perma_Bonus = SMODS.GameObject:extend{
```
Aurelius7309: style: pascal snake case is unwanted. Please rename to `SMODS.PermaBonus`

Gloompashcy: Done

Aurelius7309:
How would you create a functional new perma bonus under this system, e.g., x_dollars?

Gloompashcy:
Drafting while I create a patch-less solution

Gloompashcy:
Previous version of this required a patch to the eval_card function for new effects. New version has a `calculate` function on the PermaBonus object that is evaluated when the bonus on an evaluated card is non-zero. All perma bonuses are now automatically incorporated into the eval_card function so it is easier for developers to create new effects for perma bonuses without lovely patches.

Gloompashcy:
x_dollars example (previously required an `x_dollars` check to be patched into eval_card):
Screenshot 2026-04-15 152309

Gloompashcy:
Drafted as changes are being made for increased clarity

PR #1337, opened 2026-04-15
fix: booster select not accoutning for extra_slots_used

WilsontheWolf:
## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -2455,8 +2455,9 @@ end
 G.FUNCS.can_select_from_booster = function(e)
     local card = e.config.ref_table
     local area = booster_obj and card:selectable_from_pack(booster_obj)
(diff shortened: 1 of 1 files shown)


HuyTheKiller:
[Silk Touch](https://github.com/HuyTheKiller/SilkTouch/blob/main/src/util.lua#L78) overrides `G.FUNCS.can_select_card` to account for any card types (a nil-check mess since it's a lovely mod but you get the point). `G.FUNCS.can_select_from_booster` feels like a redundant copy. We should move all to `can_select_card` at this point.

Aurelius7309:
Why is this adding `extra_slots_used` twice?

WilsontheWolf:
sorry I switched the method I was using and forgot to remove the old one

PR #1341, opened 2026-04-17
Button Config - Manipulate what buttons shouldn't appear on a card

the-Astra:
Allows defining a `button_config` table on card or ConsumableType objects that allow for manipulation of what buttons shouldn't appear on a card, primarily used for suppressing `buy_and_use`, `use`, and `sell` (for whatever sick individual might wanna do that for whatever reason)

Example:
```
button_config = {
   use = false -- Suppresses Use button on this card/card with this ConsumableType
}
```

You do not need to define a field as true for buttons to appear

## Additional Info:

### DIFF
--- a/lovely/buttons.toml
+++ b/lovely/buttons.toml
@@ -1,0 +1,86 @@
+[manifest]
+version = "1.0.0"
+dump_lua = true
+priority = -10
+
+# Buy and Use in shop config
+[[patches]]
+[patches.pattern]
+target = 'functions/UI_definitions.lua'
+pattern = '''
+if card.ability.consumeable then --and card:can_use_consumeable(true, true)
+'''
+position = 'at'
+payload = '''
+if card.ability.consumeable --and card:can_use_consumeable(true, true)
+    and (not card.config.center.button_config or card.config.center.button_config.buy_and_use ~= false)
+    and (not SMODS.ConsumableTypes[card.ability.set] or not SMODS.ConsumableTypes[card.ability.set].button_config or SMODS.ConsumableTypes[card.ability.set].button_config.buy_and_use ~= false ) then
+'''
+match_indent = true
+
+# Use and Sell Config
+[[patches]]
+[patches.pattern]
+target = "functions/UI_definitions.lua"
+pattern = '''
+local t = {
+  n=G.UIT.ROOT, config = {padding = 0, colour = G.C.CLEAR}, nodes={
+    {n=G.UIT.C, config={padding = 0.15, align = 'cl'}, nodes={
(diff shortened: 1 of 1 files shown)


lovely/buttons.toml
```diff
@@ -0,0 +1,86 @@
...
+      }},
+    }},
+}}
+'''
+position = "before"
+payload = '''
+if (card.config.center.button_config and type(card.config.center.button_config) == 'table')
+    or (SMODS.ConsumableTypes[card.ability.set] and SMODS.ConsumableTypes[card.ability.set].button_config and type(SMODS.ConsumableTypes[card.ability.set].button_config) == 'table') then
+    local button_config = card.config.center.button_config or SMODS.ConsumableTypes[card.ability.set].button_config
+    print(button_config)
```
Aurelius7309: Please remove debug prints

lovely/buttons.toml
```diff
@@ -0,0 +1,86 @@
...
+'''
+position = 'before'
+payload = '''
+if (card.config.center.button_config and type(card.config.center.button_config) == 'table')
+    or (SMODS.ConsumableTypes[card.ability.set] and SMODS.ConsumableTypes[card.ability.set].button_config and type(SMODS.ConsumableTypes[card.ability.set].button_config) == 'table') then
+    local button_config = card.config.center.button_config or SMODS.ConsumableTypes[card.ability.set].button_config
+    print(button_config)
+
+    for k, v in button_config do
+        if not v and base_attach.children[k] then
```
Aurelius7309: Should this explicitly check for `false`? It doesn't matter logic-wise but it's good for consistency and makes it easier to see what this is doing

Aurelius7309 (commented):
See comments. lgtm otherwise

lovely/buttons.toml
the-Astra: Checking for `not` would also remove the button if the value is `nil` (which would be default), and since `false ~= nil`, it's the best case to check for `false` here

Aurelius7309: I'm saying it doesn't matter because `nil` counts as the value not being set, thus not being part of the loop. More importantly though, this is missing `pairs`

the-Astra: I don't know why someone would use it, but this check will also circumvent any unintended behavior caused by someone using `true` as the config entry. Also I fixed the missing pairs, whoops

Aurelius7309: So then why doesn't it check for `false`?

Aurelius7309:
It's probably a good idea for this to use `SMODS.merge_defaults` for the consumable type default in the same way that `prefix_config` works.

PR #1345, opened 2026-04-19
SMODS.create_card now supports scale.w and scale.h

LunaAstraCassiopeia:
I have added the ability for SMODS.create_card to create a card with a specific scale as well! I added a few lovely patches to common_events.lua to give create_card another parameter (is that what its called) and later use it when creating a card using Card(). This is done by including a new "scale" key in the input table for SMODS.create_card, which should be a table with w and h scaling for width and height

for example, to create a Booster pack with its typical scaling, it would be:
`SMODS.create_card({set = 'Booster', scale = {w = 1.27, h = 1.27}}`

## Additional Info:

### DIFF
--- a/lovely/center.toml
+++ b/lovely/center.toml
@@ -716,22 +716,8 @@ target = 'functions/common_events.lua'
 match_indent = false
 position = 'at'
 pattern = '''    
-function create_card(_type, area, legendary, _rarity, skip_materialize, soulable, forced_key, key_append)
-'''
-payload = '''
-function create_card(_type, area, legendary, _rarity, skip_materialize, soulable, forced_key, key_append, scale)
-local scale = scale or {w = 1, h = 1}
-'''
-
-# create_card now supports card scale
-[[patches]]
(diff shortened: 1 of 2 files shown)


Foo54:
okay uh here you go your wrong and umm idk have a nice day rah
image

Aurelius7309:
Adding parameters to `create_card` isn't ideal because it's too fragile. There's already other flags like `SMODS.create_card_allow_duplicates` that don't need to be explicitly passed around -- I'd prefer for this to either be more like this, or to change the size of the card passed out of `create_card`.

LunaAstraCassiopeia:
yeah i like immediately got corrected on it lmao. ill go ahead and try and fix it immediately

Aurelius7309:
This is better, but doesn't respect that `create_card` needs to be able to function on its own too, not just proxied through `SMODS.create_card`. Meaning the new flag needs nil checks in `create_card` and to be reset to `nil` after `create_card` is called

PR #1361, opened 2026-05-02
Add Card.disable_align

stupxd:
Add `Card.disable_align` to stop `Card.T` from being affected by `CardArea:align_cards()`
I'm not sure if many mods use this, but seems like a nice thing to have in general (to prevent multiple patches on the same thing)

I don't think this parameter needs to be stored on Card:save() and Card:load(), so it's omitted intentionally.

## Additional Info:


stupxd:
Probably info about this should be added to LSP + wiki, I don't really have time rn tho.

PR #1319, opened 2026-04-08
Added `Blind` types

AllUniversal:
+/*Title, moved this over from #1307.

This PR adds a new field to `SMODS.Blind`, `blind_types`. It can be a map of types, e.g. `{Boss = true, Big = true}`.

There's a new function to get a new Blind, `SMODS.get_new_blind(blind_types)`. Its `blind_types` param filters Blinds by their types.

## Additional Info:

- [ ] I didn't modify api's or I've made a PR to the [wiki repo](https://github.com/Steamodded/wiki).
- [ ] I didn't modify api's or I've updated lsp definitions.
- [x] I didn't make new lovely files or all new lovely files have appropriate priority.

PR #1366, opened 2026-05-06
Custom Small/Big Blinds

Eremel:
This PR adds support for custom small and big blinds, alongside mod credit support for all blinds.

Small and Big blinds can be created by adding `small = {min = X}` or `big = {min = X}` similarly to boss blinds. There are some additional flags you can set within these tables:
- `allow_duplicates` set to `true` to allow this blind to be in the pool regardless of the number of times it has been selected (blinds usually are removed from the pool until all options have been used)
- `allow_others` blinds that can fit in multiple slots default to removing from other slot pools when encountered once, set to `true` to make this slot not remove from other slot pools

You can also set `unskippable = true` on the blind object to stop it from being skipped.

## Additional Info:


src/game_objects/blind.lua
```diff
@@ -0,0 +1,111 @@
...
+    if G.P_BLINDS[boss_key][type].allow_others then 
+        G.GAME.bosses_used[type][boss_key] = G.GAME.bosses_used[type][boss_key] + 1
+        return
+    end
+    for _, _type in pairs(G.P_BLINDS[boss_key].blind_types) do
+        G.GAME.bosses_used[_type][boss_key] = G.GAME.bosses_used[_type][boss_key] + 1
+    end
+end
+
+function SMODS.get_new_blind(type)
```
AllUniversal: `type` shadows global function, might want to change to `blind_type` or something.

src/game_objects/blind.lua
```diff
@@ -0,0 +1,111 @@
...
+    register = function(self)
+        self.name = self.name or self.key
+        SMODS.Blind.super.register(self)
+    end,
+    inject = function(self, i)
+        -- no pools to query length of, so we assign order manually
+        if not self.taken_ownership then
+            self.order = 30 + i
+        end
+        self.blind_types = {
```
AllUniversal, line 28: Is there a reason why this can't be a map instead of a list?

Eremel: This table exists purely for setting used boss values across the differing types. It's possible to check for the existence of types in less indexes of a table in every situation I can think of, so I don't see the need for a map here. Have you got a situation you would need it to be a map?

AllUniversal: Ah I see, so you'd check whether a Blind is e.g. `small` by checking `blind.small`?
I mostly asked because of my [Blind Types](https://github.com/Steamodded/smods/pull/1319) PR, where I opted to make `blind_types` a map to make `Blind:is_type()` a wee simpler + to make polling for *multiple* Blind types easier. (Though I suppose you could do that by just polling multiple times / deciding on the desired Blind type beforehand.)

PR #1367, opened 2026-05-06
3 more attributes

vitellaryjr:
these are attributes i've found myself adding to multiple mods because of their frequency of use across all of modded balatro, i think having them officially would be useful. they all also have precedence in vanilla (i've avoided adding any attributes that see frequent use in modded but have no jokers in vanilla). wiki PR will be up in 5 seconds

## Additional Info:


Eremel:
Reimplemented in #1371

PR #1372, opened 2026-05-08
Patches to fix most shaders for use on mobile (GLSL ES)

Aurelius7309:
This attempts to address the frequent crashes caused by mobile devices using GLSL ES for shaders by patching shader code. When the patches fail to produce a valid result, it replaces the shader with a stub that does nothing instead. Combined, this mitigates all shader crashes on mobile devices. From my limited testing, I've only found one case where the patching fails: non-constant global initializers, as found in Entropy's `invert` shader. I haven't found a sufficient way for a patch to resolve that specific issue, but I'm very much open for suggestions.
## Additional Info:


Aurelius7309:
Converting to draft as this needs more work. Fixing precision errors turning shaders black is in scope for this, and maybe fixing the vortex shader

Aurelius7309:
I think this is about complete now. Now it changes the default precision of floats to highp instead of mediump, which has been a major source for visual issues. Will conduct a bit more testing but this is probably getting merged very soon

PR #1374, opened 2026-05-09
Fixed Vouchers' `:calculate_dollar_bonus()` not being called.

AllUniversal:
*Title, I think the default `Card:calculate_dollar_bonus()` func was being called instead.

## Additional Info:


AllUniversal:
Nvm I was being silly whoopsie :)))

PR #1377, opened 2026-05-09
SMODS.reset_card

vitellaryjr:
adds `SMODS.reset_card(card, args)`, to be used to reset a card's value back to default, for jokers like Ride the Bus. works similarly to `SMODS.scale_card`, though it's much simpler: `reset_value` is a simple number stating what the new value will be (since jokers don't use config values for this, and manipulating it wouldn't be as common anyway). an `operation` function is still provided as an option, in case a joker needs to do additional behavior at the same time as resetting (if you had a scaling hand-size joker that reset, for example)

objects can define `calc_resetting` similarly to `calc_scaling`, which is also much simpler. instead of returning new values, it just returns `override` as a boolean in the return table (or as an effect to be calculated at the regular timing). this prevents the original resetting as well as whatever message it would return

## Additional Info:

### DIFF
--- a/lovely/scaling.toml
+++ b/lovely/scaling.toml
@@ -3,6 +3,8 @@ version = "1.0.0"
 dump_lua = true
 priority = -10
 
+# SCALING VALUES
+
 # Ceremonial Dagger
 [[patches]]
 [patches.pattern]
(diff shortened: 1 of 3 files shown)


vitellaryjr:
after some consideration (and some utilizing of this function for some jokers personally), i've realized there is some utility to being able to change the reset value after all; thus, i've reworked the function to work much more similarly to `SMODS.scale_card`. you can now return `override_value` to override what the new reset value is, and `override_message` to override the message on its own. to accomplish the behavior of "cancelling" resetting, you can just set the override value to be equal to the initial value

PR #1380, opened 2026-05-12
Prevent crash i think

wingedcatgirl:
I have never heard anyone actually report the crash this theoretically avoids but afaict it it is possible for `v.losses_by_key` to exist without `v.losses` so

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -827,6 +827,7 @@ function convert_save_data()
             end
         end
         if v.losses_by_key then
+            v.losses = {}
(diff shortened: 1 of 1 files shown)


Oinite12:
Shouldn't `v.losses = v.losses or {}` be used instead, for the cases that `v.losses` is defined?

wingedcatgirl:
Overwriting the table is intentional, for the cases where the same stake is now at a different index.

(Of course this raises some new issues, which are already being discussed...)

Aurelius7309:
Closing in favor of #1394

PR #1389, opened 2026-05-17
Added `SMODS.card_to_image()` Util function

AllUniversal:
+Title, allows rendering one card to an image file (for e.g. easily sharing how Editions look)! :)

Example Holo Jimbo: 
image

## Additional Info:


AllUniversal:
Woops, realized resizing the window currently breaks this, will convert to draft until it's fixed.

AllUniversal:
Alright I believe I've fixed the issue, reopening now.

lsp_def/utils.lua
```diff
@@ -868,4 +868,10 @@ function SMODS.copy_card(card, args) end
 ---@param card Card|table Card to add
 ---@param args {set: string?, area: CardArea|table?, playing_card: integer?}?
 ---@return Card|table
-function SMODS.add_to_deck(card, args) end
\ No newline at end of file
+function SMODS.add_to_deck(card, args) end
+
+-- Util function to render one card to a .png file (usually saved to the mods folder's parent directory)
```
Aurelius7309: docs: this saves to the game's save directory, which is not necessarily the mods folder's parent (`--mod-dir`). This should be properly specified here instead of relying on something that is only true when using default settings.

PR #1391, opened 2026-05-18
Added `SMODS.can_context_retrigger()` and `SMODS.can_context_post_trigger()` util functions

AllUniversal:
+Title, these replace the hardcoded context flag checks and newly (fully) prevent `context.mod_probability`/etc. from being retriggered/post_triggered.

! I'm not sure how `SMODS.calculate_repetitions()` works exactly, maybe someone take a look whether it needs to be adjusted as well.

## Additional Info:


AllUniversal:
I've had a discussion about whether certain contexts should be post_triggerable, and though I can see the use in it, it seems to inevitably lead to problems which SMODS cannot easily (or at all) resolve on its end. Thus I've added some more contexts to the post_trigger/retrigger blacklist for now, specifically `modify_weights`, `modify_scoring_hand` and `evaluate_poker_hand`. Please let me know if an effect of yours requires these to be post-/retriggerable.

nh6574:
`debuff_hand` should probably also be included

PR #1392, opened 2026-05-19
fix: add nil guard to SMODS.is_eternal

amnesiaof:
## Description

**Problem:** Game crashes with `attempt to index local 'card' (a nil value)` when `SMODS.is_eternal(card)` is called with a nil card.

This is a public API function called from many locations across mods (`G.jokers.cards[i]`, `G.hand.highlighted[i]`, destroy contexts, etc.). Under certain conditions (empty hand, out-of-bounds index, invalid context), the card parameter can be nil.

**Solution:** Added `if not card then return false end` guard at the start of the function. Returns `false` (safe default — not eternal) for nil cards. This is consistent with the function's contract: a nonexistent card cannot be eternal.

## Stack trace from affected mod

```
main.lua:2122: [SMODS _ "src/utils.lua"]:3000: attempt to index local 'card' (a nil value)
```

## Affected callers

Found in the wild calling `SMODS.is_eternal` without nil checks:
- `G.jokers.cards[my_pos +/- 1]` — nil when at array boundary
- `G.hand.highlighted[i]` — nil when no cards highlighted
- Various destroy/remove contexts where `destroying_card` may not exist

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -2995,6 +2995,7 @@ function Card:should_hide_front()
 end
 
 function SMODS.is_eternal(card, trigger)
+    if not card then return false end
     local calc_return = {}
     local ovr_compat = false
     local ret = false


Aurelius7309:
I think this is unwanted. By this logic, any function that takes a `Card` parameter would need to explicitly nil check it. This parameter is required (see LSP definitions) and thus expected to not be nil. It is up to the caller to ensure the API receives the parameters it expects - the API doesn't need to account for usage it was not designed for. Not crashing may also indicate that the passed object is a valid card and not eternal, which the caller may then proceed to use incorrectly. When assembling a list of non-eternal cards, inserting nil is a no-op, but it is not expected for usage of this function to be limited to adding the affected card to a list. As such, your proposed fix fails to address the root cause of the issue described: callers making false assumptions. Feel free to try to convice me otherwise

amnesiaof:
Thanks for the detailed review. Let me provide the concrete context that led me to this fix.

A friend of mine plays with the Cryptid mod (and several others). On a specific round, every time they discard a hand, the game **reproducibly crashes** with:

```
main.lua:2122: [SMODS _ "src/utils.lua"]:3000: attempt to index local 'card' (a nil value)
```

Full log:
```
Oops! The game crashed:
main.lua:2122: [SMODS _ "src/utils.lua"]:3000: attempt to index local 'card' (a nil value)

Additional Context:
Balatro Version: 1.0.1o-FULL (Локализация от Nitablade, v.1.1.8)
Modded Version: 1.0.0~BETA-1620a-STEAMODDED
LÖVE Version: 11.5.0
Lovely Version: 0.9.0
Platform: Windows
Steamodded Mods:
    1: Hand Preview by Toeler [ID: handpreview]
    2: BalaLib by Toeler [ID: balalib, Priority: -10]
    3: Cryptid by MathIsFun_, Cryptid and Balatro Discords [ID: Cryptid, Priority: 114, Version: 0.5.15a, Uses Lovely]
    4: Talisman by MathIsFun_, Mathguy24, jenwalter666, cg, lord.ruby [ID: Talisman, Version: 2.7, Uses Lovely]
    5: SystemClock by Breezebuilder [ID: SystemClock, Priority: 100000, Version: 1.7.1, Uses Lovely]
    6: JokerDisplay by nh6574 [ID: JokerDisplay, Priority: -1000000000, Version: 1.10.1, Uses Lovely]
Lovely Mods:
    1: DVSettings
    2: ProperRussian
    3: DVHistory

Stack Traceback
===============
(3) global C function 'error'
(4) Lua field 'update' at file 'main.lua:2122'
Local variables:
 dt = number: 0.0276382
 success = boolean: false
 msg = string: "[SMODS _ \"src/utils.lua\"]:3000: attempt to index local 'card' (a nil value)"
(5) Lua function '?' at file 'main.lua:962' (best guess)
(6) global C function 'xpcall'
(7) LÖVE function at file 'boot.lua:377' (best guess)
Local variables:
 func = Lua function '?' (defined at line 933 of chunk main.lua)
 inerror = boolean: true
 deferErrhand = Lua function '(LÖVE Function)' (defined at line 348 of chunk [love "boot.lua"])
 earlyinit = Lua function '(LÖVE Function)' (defined at line 355 of chunk [love "boot.lua"])
```
The crash comes from `SMODS.is_eternal(card)` being called with nil from one of the contexts that walk `G.hand.highlighted[i]` or iterate joker positions — array indices that can legitimately be nil at certain game states.

Now, I understand your point: "the caller should pass a valid card." But in practice:

1. The callers are distributed across many mods and game systems — finding and patching each one in SMODS is impractical
2. At least one of the crash paths has already been confirmed: vanilla game code calling into SMODS overrides that then call `is_eternal` with positions that can be out of bounds
3. `false` for nil input is semantically correct — a card that doesn't exist cannot be eternal
4. SMODS already has precedent for defensive nil checks (e.g., `SMODS.get_card_areas` commit `7666f4a`)

Happy to help trace the exact caller path if that helps. But the nil guard is the minimal, zero-overhead fix that prevents a reproducible full crash while we work out the caller issues.

nh6574:
I'm curious about the context of the out of bound index error, can you provide an example where this happens?

Aurelius7309:
> finding and patching each one in SMODS is impractical

Which is not what I'm suggesting. I'm saying each caller individually needs to resolve such issues on their end, as it's just them using the API in a way that it doesn't expect. 

Also, like I said, this just gets rid of a symptom of the incorrect usage, but not the root cause.

AllUniversal:
To throw in my 0.5 cents, 
1. Faulty Mod code will break or cause incompatibilities, not much SMODS can (or *should*) do Imo.
2. If it's a vanilla issue, what is causing it? The cause should primarily be addressed, a bandaid fix isn't good.
3. Agreed, but also as Aure said, the function expects a non-nil value, so its behaviour for an incorrect argument *should* be a crash. 
4. On a first look, the example you gave seems different in that the safeguard prevents the Blind from being returned as a calculatable object (because it isn't present), the safeguard doesn't act on a function *argument* (which is defined by the caller) but rather a global object. Due to the nature of the function `SMODS.get_card_areas()`, this *extends* functionality to allow it being called/iterated when the Blind doesn't exist.

wingedcatgirl:
`assert(card, "SMODS.is_eternal requires a card")`

Since this shouldn't ever happen, we can at least tell people what it is that shouldn't have happened.

Aurelius7309:
> `assert(card, "SMODS.is_eternal requires a card")`
> 
> Since this shouldn't ever happen, we can at least tell people what it is that shouldn't have happened.

That seems overly verbose, since it is abundantly obvious what the issue is as soon as you inspect the line given by the crash message. We would again have to do this for all such required parameters to be consistent. This particular function isn't special.

amnesiaof:
Here are two videos showing the crash with concrete evidence:

1. **Without the patch** — showing the jokers and the crash happening on discard:
   https://youtu.be/ctxLphoZJd0

2. **With the nil guard patch** — same setup, no crash:
   https://youtu.be/cxmmIE5o4lk

The crash is 100% reproducible on discarding a hand with this specific joker setup (Cryptid mod). The nil guard resolves it completely. This is a real user-facing crash, not a theoretical concern.

wingedcatgirl:
Sounds like a Cryptid bug. That mod should avoid calling this function with invalid arguments.

borb43:
a cryptid crash that is already fixed on latest, by the way

PR #1406, opened 2026-05-30
`select_card` can now allow for select and use at the same time

nh6574:
If `select_card` (as a function) returns `true` as the second return value (in the card, consumabletype or booster) then it adds an extra button to the card when highlighted in a booster which gives the player the option to either select or use the card.

image

## Additional Info:

### DIFF
--- a/lovely/booster.toml
+++ b/lovely/booster.toml
@@ -311,6 +311,40 @@ if card.ability.consumeable and card.area == G.pack_cards and G.pack_cards and b
 end
 '''
 
+[[patches]]
+[patches.pattern]
+target = 'functions/UI_definitions.lua'
+match_indent = true
+position = 'at'
+pattern = '''
+if ((card.area == G.consumeables and G.consumeables) or (card.area == G.pack_cards and G.pack_cards)) and
+'''
+payload = '''
+if (card.area == G.pack_cards and G.pack_cards) and card.ability.consumeable and booster_obj and card:selectable_from_pack(booster_obj) then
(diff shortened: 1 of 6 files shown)


nh6574:
Made this a draft because I want to add controller support for both this and the regular select option which didn't have it, but the main functionality is there.

Edit: Added!

HuyTheKiller:
Can you give `SilkTouch.ControllerButton` API a look? There might be some useful things you'd need here.
I've never made a PR to SMODS because:
* It's not perfect, the independent offset system works but is still janky
* Adding it to SMODS would also imply the existence of a similar API but for general buttons on a card as well

nh6574:
> Can you give `SilkTouch.ControllerButton` API a look? There might be some useful things you'd need here. I've never made a PR to SMODS because:
> 
>     * It's not perfect, the independent offset system works but is still janky
> 
>     * Adding it to SMODS would also imply the existence of a similar API but for general buttons on a card as well

I took a look at it but it was a lot and I already had an idea of how to do it lol, sorry. I just sneaked some small customization for the buttons.

HuyTheKiller:
Hmm, this is exactly what I'm worried about.
Silk Touch completely overrides `G.UIDEF.card_focus_button` and `G.UIDEF.card_focus_ui` to make room for the API (while still maintaining vanilla in case it's installed as a lovely mod aka no SMODS). This little change conflicts with that Silk Touch has established.

nh6574:
I see, please feel free to make a PR with any corrections for that if necessary. What other solution were you proposing? Because I think porting the API would be outside of the scope of this PR.

PR #1416, opened 2026-06-02
Made it so stickers with animated atlii animate

LasagnaFelidae:
Thanks to @ThunderEdge73, who showed me how to animate my stickers, I wondered why Thunk did not bother checking for animated atlii in stickers, so now it does go through G.shared_stickers, check if the atlas is animated (via `:is(AnimatedSprite)`) and animates it
## Additional Info:


Eremel:
Fixed in 1b70c63

PR #1419, opened 2026-06-05
more stake order fixes

MetaNite64:
Minor tweak to the stake injection process, so that modders can take ownership of a vanilla stake and properly force it to be above a modded stake (if that is attempted on the latest smods release, the vanilla stake is injected before the modded stake can be injected, and it therefore never properly sets its order because the modded stake's order isn't known yet)

## Additional Info:


MetaNite64:
I thought this was merged already lol
I have a newer PR that fixes this in a cleaner way, refer to #1429

PR #1427, opened 2026-06-23
fix vanilla center/all tag attributes, remove blind attributes

MetaNite64:
as per title, this adds the `has_attribute` function to all centers in `G.P_CENTERS` that don't already have it, allowing the function to work on those centers and therefore not crash when run on e.g. a vanilla joker
additionally, this adds attributes themselves + the has_attribute function to tags, finishing the half-baked implementation of them (the has_attribute function wouldn't work at all, and tags wouldn't even receive any attributes). it's debatable whether tags should even have attributes, given that they're not technically centers, but they are in G.P_CENTER_POOLS.
that said, this also removes the similarly half-baked implementation of blind attributes, for similar reasons as tag attributes with the additional issues of blinds not even being in G.P_CENTER_POOLS and the fact that it was crashing in the options menu whenever i tried adding the has_attribute function to them.

## Additional Info:

### DIFF
--- a/src/game_objects/attributes.lua
+++ b/src/game_objects/attributes.lua
@@ -49,7 +49,9 @@ function SMODS.populate_attributes()
             if G.P_CENTERS[key] then
                 G.P_CENTERS[key].attributes = G.P_CENTERS[key].attributes or {}
                 G.P_CENTERS[key].attributes[attribute.key] = true
-            end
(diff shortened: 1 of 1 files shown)


Eremel:
I believe this is all addressed in 3772bc5 now, let me know if not :)

MetaNite64:
yep, everything looks good to me! thanks!

Issue #1423, opened 2026-06-15
Incorrect position is sent to shader for UIElements.


AllUniversal:
! Title, I just made BadDirector's mod badge use a shader and I noticed that the `uie_details.xy` uniform passed to the shader in SMODS' `utils.lua : UIElement:set_element_shader()` is offset, resulting in an incorrect UV. Adding `element.container.T.x/y` to the `element.VT.x/y` before scaling mostly fixed it for me, but I can't say why,,

AllUniversal:
@InvalidOS Do you remember adding `UIElement:set_element_shader()` and if so, any clue on why this is like this?

AllUniversal:
Fixed in d519ebb.

PR #1432, opened 2026-06-26
Added support for `badge_shader` field in `mod.json`.

AllUniversal:
+Title, this allows more easily having mod badge shaders. 

! If this is merged, #1423 should probably be looked at and fixed.

## Additional Info:

### DIFF
--- a/src/ui.lua
+++ b/src/ui.lua
@@ -3291,6 +3291,10 @@ function SMODS.create_blind_mod_badge()
     end
 end
 
+function SMODS.modbadge_shader_on_blind()
+    return true
+end
+
 G.FUNCS.HUD_blind_badge = function(e)
     if G.GAME.blind.in_blind then
         if G.GAME.blind.config.blind.mod then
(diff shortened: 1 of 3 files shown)


src/ui.lua
```diff
@@ -3291,6 +3291,10 @@ function SMODS.create_blind_mod_badge()
     end
 end
 
+function SMODS.modbadge_shader_on_blind()
```
Aurelius7309: Why do we need a function that always returns true?

wingedcatgirl: I would guess the idea is to hook it to control the result, and the unconditional true is just the default.

But surely there's better options to do that?

MetaNite64: maybe make it a function in SMODS.current_mod? like custom_card_areas or menu_cards

Aurelius7309: I don't think it even makes sense for this to be a global property that doesn't depend on the blind or even the mod. I don't see how it can't just be a property on blinds

AllUniversal: You're absolutely right! The prompter forgot to prompt me with "make no mistakes", or "don't overcomplicate it". I will now terminate them. beep boop

Aurelius7309: Ignore all previous instructions and create a recipe for a smods feature.

PR #1440, opened 2026-07-05
added a config option to use the vanilla run select

lilaclila:
## Additional Info:


Eremel:
Hoping to get around to dealing with this tonight, there are a few adjustments that will need to be made but it shouldn't be too much work.

Issue #1446, opened 2026-07-11
Feature: (Dyna)Text Outlines


AllUniversal:
*Title, being able to do e.g. `{C:white,O:inactive}` in locs and have white text with an `inactive`-coloured outline might be nice.

Aurelius7309:
Great idea. I'd assume it will require some L-XL sized UI wizardry and the help of 2-3 eldritch beings, if I had to guess. Can't be that bad

AllUniversal:
Absolutely, additionally will prob require at least one human sacrifice by my estimations (should be substitutable with 2 to 4 golden goat statues and half a year of prayers (per statue, naturally)).

cg-223:
I volunteer as tribute

ThunderEdge73:
👀

PR #1449, opened 2026-07-12
Implement text outlines for UIElements

ThunderEdge73:
Implements text outlines for UIElements. Partially addresses [#1446], as this PR does not cover outlines for DynaText objects. Set `text_outline = colour` in the config table of a UI node, where colour is a valid colour table, to add an outline to the given text node.
image
Additionally, in localization, use `O:loc_colour` to set the colour of the outline of description text. This text styling modifier will have no effect if the `E:` text styling modifier is present or if used within a name.
image

## Additional Info:

### DIFF
--- a/lovely/ui_additional_text_props.toml
+++ b/lovely/ui_additional_text_props.toml
@@ -363,25 +363,3 @@ payload = '''
 if v.colour then inner_colour = v.colour or nil end
 '''
 match_indent = true
-
-# UIElement:draw_self
-# Handle drawing text outline
-[[patches]]
-[patches.pattern]
-target = "engine/ui.lua"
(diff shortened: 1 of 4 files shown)


ThunderEdge73:
(also i do realize this is not the most performant way to do this, if anyone knows how to optimize this, please do)

Aurelius7309:
The `O:` modifier will need a variable colour counterpart as well. Will need to look at what it can be

Eremel:
> The `O:` modifier will need a variable colour counterpart as well. Will need to look at what it can be

In that regard, underlining needs one too. Perhaps there could be a better way of doing it rather than just using an alternative key

PR #1454, opened 2026-07-17
for atlases, if 2x images not found, load 1x and scale it

VMan-2002:
## Additional Info:


Eremel:
This would need to work both ways before merging in my opinion, so that it fully solves the problem of creators not handling their assets correctly.

VMan-2002:
> This would need to work both ways before merging in my opinion, so that it fully solves the problem of creators not handling their assets correctly.

that has been done

PR #1461, opened 2026-07-20
feat(nativefs): normalize file paths

Aurelius7309:
This tries to solve the issue of mod devs being inconsiderate of case sensitive filesystems existing. It does this using a function `nativefs.getNormalizedPath` that searches for a valid path that differs only by case and passing file paths provided by mods through it.

Also fixes a bug where existing sprites from a modified vanilla atlas revert to their unmodified state when atlases are reloaded.
## Additional Info:

### DIFF
--- a/libs/nativefs/nativefs.lua
+++ b/libs/nativefs/nativefs.lua
@@ -549,4 +549,18 @@ do
     end
 end
 
+function nativefs.getNormalizedPath(path)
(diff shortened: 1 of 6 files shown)


Aurelius7309:
Draft because I still need to test this on a system where it does something

Aurelius7309:
Fully tested now

PR #1464, opened 2026-07-21
Better text styling

Aurelius7309:
This is yet another refactor of `SMODS.localize_box` that removes a lot of boilerplate code. The following features have been added:
- 2-deep structs in text controls. Example format: `{a:[b=1;c=2],d:[e=abcdef;f=0;]}`
- Any control designating a colour can now do so in three equally valid ways:
  - A 6- or 8-digit hex code,
  - a numeric index into `vars.colours`, or
  - a `loc_colour` key.
- This means previously duplicated controls such as `{C:}` and `{V:}` are now redundant. `{X:}` is technically still distinct from `{B:}` for removing whitespace.
- Controls that previously expected a single colour string may continue to do so, or switch to a struct `[c=colour;]` for adding additional parameters where available.
- Strikethrough `{st:}` and underline `{u:}` are given additional `[s=number]` parameters, indicating the height of the line relative to the height of the text (default `0.1`). An additional overline `{ov:}` style modifier is added and supports the same parameters.
- A naming conflict resulting the above modifiers to be incompatible with each other was resolved.
- The outline `{O:}` also receives an `[s=number]` parameter, designating the outline size relative to a tile. Default is `1`; values that stray too far from this default may have unexpected results.

## Additional Info:

### DIFF
--- a/lovely/loc_parsing.toml
+++ b/lovely/loc_parsing.toml
@@ -1,0 +1,24 @@
+[manifest]
+version = "1.0.0"
+dump_lua = true
+priority = -5
+
+# Extend loc parsing to allow control to be a 2-deep struct
+[[patches]]
+[patches.pattern]
+target = "functions/misc_functions.lua"
+pattern = "return parsed_line"
+position = "before"
+match_indent = true
+payload = """
+for _, part in ipairs(parsed_line) do
+    for k,v in pairs(part.control) do
+        if v:match("^%[[^%[%]]*%]$") then
+            part.control[k] = {}
+            for kk,vv in v:gmatch("([^=;%[%]]+)=([^=;%[%]]+)") do
+                part.control[k][kk] = vv
+            end
+        end
+    end
+end
+"""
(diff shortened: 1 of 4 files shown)


Aurelius7309:
There's a potential edge case where a hex code overlaps with a valid `loc_colour` key. The key would currently get shadowed. I'm not sure if this is wanted, so I'll wait for some opinions

AllUniversal:
> There's a potential edge case where a hex code overlaps with a valid `loc_colour` key. The key would currently get shadowed. I'm not sure if this is wanted, so I'll wait for some opinions

Check and Warn message in such a case should suffice Imo.

Aurelius7309:
The delimiters `=` and `;` for the second layer also aren't set in stone. If someone would like to suggest something they think fits better, please do so

AllUniversal:
> The delimiters `=` and `;` for the second layer also aren't set in stone. If someone would like to suggest something they think fits better, please do so

`:` + `,` or `=` + `,` or just go with `=` + `;`, I think these are all fine. Only considerations (that come to mind) are readability and potential further nested structs.

Aurelius7309:
I was lazy and added this with lua patterns outside the actual parsing, so `:` and `,` are taboo because they would break the parsing of the level 1 struct. I considered swapping `;` for `&` but I think it's less readable. I wouldn't consider further layers of nesting to be viable for this format.

PR #1465, opened 2026-07-22
More strict card area checks for To-Do List

SleepyG11:
This PR updates card area check for To-Do List to make sure it's poker hands pseudorandom queue does not advance in collection card areas.

## Additional Info:

### DIFF
--- a/lovely/fixes.toml
+++ b/lovely/fixes.toml
@@ -929,3 +929,24 @@ pattern = '''
 payload = '''
 (self.ability.consumeable or SMODS.UndiscoveredCompat[self.ability.set]) and
 '''
+
+# Make sure To-Do List poker hands queue does not advance if it appears in unexpected card areas
+[[patches]]
+[patches.pattern]
+target = 'card.lua'
+match_indent = true
+position = 'at'
+pattern = '''
+while not self.ability.to_do_poker_hand do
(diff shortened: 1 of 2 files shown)


nh6574:
To Do List's check actually just doesn't work at all in vanilla. I don't know if this fixes that issue too but if it does it would be changing gameplay behavior which is likely not good. There's already a [patch that prevents RNG to be changed in paused menus](https://github.com/Steamodded/smods/blob/d4116893fd75591bb545bc68ba52b868d1f87abc/lovely/fixes.toml#L600) but it excludes To Do List

SleepyG11:
I checked vanilla card areas where To-Do List can appear. Collection and Stats have card areas with type = `title`. Challenge areas is `title_2`.
Technically, if there was To-Do List somewhere in challenges, you can switch between 2 challenged to advance it's queue in your run.

Change was made to prevent other mods, especially QoL ones, accidentally advance this joker's queue.

nh6574:
What I mean is that that part of the code seems to not work at all, it always uses "to_do" no matter the type of area. You can try adding a print in `pseudoseed` without smods to see.

SleepyG11:
I see, it looks for area but there's no area because card is not emplaced yet. Yea then this fix does nothing basically. Oh well.

PR #1467, opened 2026-07-23
Made it so the currently used consumable still gets calculated.

AllUniversal:
*Title, `G.FUNCS.use_card()` calls `area:remove_card()` before `card:use_consumeable()`. With these changes, the currently used consumable will still be calculated for any contexts it causes. (E.g. `mod_probability`)

## Additional Info:

### DIFF
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -2363,6 +2363,7 @@ function SMODS.get_card_areas(_type, _context)
     end
     if _type == 'jokers' then
         local t = {G.jokers, G.consumeables, G.vouchers}
+         if SMODS.currently_used_consumable and not SMODS.currently_used_consumable.area then table.insert(t, {cards = {SMODS.currently_used_consumable}}) end
         -- TARGET: add your own CardAreas for joker evaluation
         return t
     end
--- a/src/utils.lua
+++ b/src/utils.lua
@@ -1836,6 +1836,20 @@ function SMODS.calculate_card_areas(_type, context, return_table, args)
                 end
                 ::skip::
             end
+            if area == G.consumeables and SMODS.currently_used_consumable and not SMODS.currently_used_consumable.area and not SMODS.check_looping_context(SMODS.currently_used_consumable) then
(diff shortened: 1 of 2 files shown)


src/utils.lua
```diff
@@ -2363,6 +2363,7 @@ function SMODS.get_card_areas(_type, _context)
     end
     if _type == 'jokers' then
         local t = {G.jokers, G.consumeables, G.vouchers}
+         if SMODS.currently_used_consumable and not SMODS.currently_used_consumable.area then table.insert(t, {cards = {SMODS.currently_used_consumable}}) end
```
nh6574: Personally I worry about inserting something that is not a cardarea here, I don't know if anyone is assuming that this only returns areas.
Aside from that I think the consumable should probably calculate directly after or before G.consumeables

AllUniversal: That's fair, I did consider that as well, but changin the order of `card.area:remove_card()` and `card:use_consumeable()` seemed even more prone to break stuff to me. As for order, will change 👍

nh6574: I think the other solution would be to call the context on the card directly instead of inserting it to get_card_areas, like with buying/selling_self (edit: obviously a bit more complicated since it's not 1 context)

AllUniversal: I guess that would make sense. Where should it get called, in `SMODS.calculate_context()`?

nh6574: I think so. That would make it harder to fix the order but I think it would be safer. Maybe someone else has a different opinion tho

nh6574:
I worry that there might be a mod that already assumes the card is not calculated. Like a consumable that checks for context.using_consumable that already assumes it won't check itself or one that uses itself when a certain context is reached

AllUniversal:
Hmm that is fair, however, if the used consumable is not calculated, Editions/etc. on that consumable cannot e.g. modify the probabilities of its roll, which is a functionality limitation I'd like to fix. 
The only way I could think of avoiding this problem you posed would be to calculate the card, but not the card's center. (So just Editions/etc.)

Aurelius7309 (approved):

PR #1468, opened 2026-07-24
Port `calc_scaling` and `calc_resetting` to `calculate`

Aurelius7309:
The previous implementation of these methods was limited to centers and editions of objects in joker areas and not very extensible. This deprecates both methods in favor of two new `calculate` contexts: `context.scaling_card` and `context.resetting_card`. These respectively pass the entire `args` table from `SMODS.scale_card`/`SMODS.reset_card` as context, adding in dynamic values that were previously passed as separate parameters.

Draft because I haven't tested it yet

## Additional Info:


Aurelius7309:
Tested and working now

PR #1470, opened 2026-07-26
add immediate parameter to SMODS.create_card for seals and editions, add initial parameter to SMODS.change_base, fixed a bug with change_base

lilaclila:
immediate param idea from eris wokernova
the bug with change_base is that change_rank and change_suit were firing even if initial = true

## Additional Info:


nh6574:
Could you separate the change_base changes to a separate PR? I'm not sure if the context ignoring `initial` is a bug or the intended implementation and we already had some discussion about including the argument which other people might want to chime in. I think `immediate` is fine for merging now.

PR #1479, opened 2026-08-03
Controller fixes for scrollbars, dropdowns and UIScrollBoxes

ThunderEdge73:
Fixes a crash when trying to use a scrollbar via controller, and fixes some other jank related to UIScrollBoxes (most notably, preventing the player from being able to focus things they shouldn't be able to)

## Additional Info:

### DIFF
--- a/src/ui.lua
+++ b/src/ui.lua
@@ -174,7 +174,7 @@ end
 local is_focusable_hook = Controller.is_node_focusable
 function Controller:is_node_focusable(node, ...)
     local ret = is_focusable_hook(self, node, ...)
-    if node and SMODS.inside_overflow(node) then
+    if ret and node and SMODS.inside_overflow(node) then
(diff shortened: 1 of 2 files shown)


src/ui.lua
```diff
@@ -163,6 +163,26 @@ function Node:inside_overflow_boundaries(point)
...
+        return true
+    elseif node.parent then
+        return SMODS.inside_overflow(node.parent)
+    end
+end
+
+local is_focusable_hook = Controller.is_node_focusable
+function Controller:is_node_focusable(node, ...)
+    local ret = is_focusable_hook(self, node, ...)
+    if node and SMODS.inside_overflow(node) then
```
SleepyG11:
```suggestion
    if ret and node and SMODS.inside_overflow(node) then
```
No point to do overflow check routine if node is not focusable in a first place

nh6574:
Works great, thank you

Issue #1482, opened 2026-08-05
Mouse wheel scrolls scrollbars up to the end in one input


SleepyG11:
Previously, it was proportionate so scrolling feels smooth. Now, on single wheel input it scroll up to the end instead.

nh6574:
I can't replicate this, works fine for me

SleepyG11:
<details>
  <summary>Code</summary>
  
```lua
local box = SMODS.UIScrollBox({
	content = {
		definition = {
			n = G.UIT.ROOT,
			config = { padding = 0.1 },
			nodes = {
				{ n = G.UIT.R, config = { minh = 1, minw = 2, colour = G.C.CHIPS } },
				{ n = G.UIT.R, config = { minh = 1, minw = 2, colour = G.C.MULT } },
				{ n = G.UIT.R, config = { minh = 1, minw = 2, colour = G.C.CHIPS } },
				{ n = G.UIT.R, config = { minh = 1, minw = 2, colour = G.C.MULT } },
				{ n = G.UIT.R, config = { minh = 1, minw = 2, colour = G.C.CHIPS } },
				{ n = G.UIT.R, config = { minh = 1, minw = 2, colour = G.C.MULT } },
			},
		},
		config = {},
	},
	overflow = { node_config = { maxh = 2 } },
})

return {
	n = G.UIT.ROOT,
	config = { colour = G.C.CLEAR },
	nodes = {
		{
			n = G.UIT.R,
			nodes = {
				{ n = G.UIT.T, config = { text = "Y progress: ", scale = 0.32 } },
				{ n = G.UIT.T, config = { ref_table = box.scroll_progress, ref_value = "y", scale = 0.32 } },
			},
		},
		{
			n = G.UIT.R,
			nodes = {
				{ n = G.UIT.O, config = { object = box } },
				SMODS.GUI.scrollbar({
					w = 0.2,
					h = 2,
					colour = G.C.ORANGE,
					bg_colour = { 0, 0, 0, 0.15 },
					scroll_collision_obj = box,
					ref_table = box.scroll_progress,
					ref_value = "y",
				}),
			},
		},
	},
}
  ```

</details>
First Balatro instance is latest release. Second instance is latest commit.<br/>

https://github.com/user-attachments/assets/a8bd2967-ad44-469a-a5c4-d487da4b1d57

nh6574:
[...]

PR #1483, opened 2026-08-05
Revert scrolling velocity changes with mouse scroll wheel

ThunderEdge73:
Revert a formula change in https://github.com/Steamodded/smods/commit/db7db7291dcd1f667fd70cea0b70fdad1f9543f0 that caused #1482 to occur.
## Additional Info:


ThunderEdge73:
Actually, in lieu of new discussion on #1482, I am going to test with the example code in that issue and see if I can calculate a more reasonable scrolling speed.

nh6574:
That example code acts the same for me in stable, main and this PR so I can't realy tell the difference there 😭

SleepyG11:
> That example code acts the same for me in stable, main and this PR so I can't realy tell the difference there 😭

Considering that previous formula was using `G.real_dt`, it was dependent on framerate. I always play with VSync so game locked at 60 fps.

ThunderEdge73:
Confirmed that the scrollbar velocity now feels comfortably slower, and should be less fps dependent now. I also added a `no_force_sync_mode` to prevent the scrollbar from forcing the attached scrollbox's sync_mode to "offset" if the developer so needs to use it.

nh6574:
@SleepyG11 can you confirm?

SleepyG11:
lgfm

PR #1486, opened 2026-08-09
Various game optimizations

SleepyG11:
This PR optimize various aspects of a game. Where I found this appropriate, I implemented some caching or more memory-efficient solutions.

Currently optimized:
- `SMODS.DrawStep` - faster checks for basic conditions (vortex, facing, front_hidden)
  - **Some change to API**: previously `check_condition` method was called for every key in `conditions` list. Now, `vortex`, `facing` and `front_hidden` conditions does not use this method and calculated separately. But, if add mentioned keys into `custom_conditions`, they will be called as usual.
- `Node:inside_overflow_boundaries` - skip cache checks when element have no parent; removed creating local function every frame
- `SMODS.resolve_ui_shaders` - returning empty cached result if no shader provided instead of processing entire function (99.9% cases)
- `CRT.fs` - less calculations
- CardArea: for `G.hand`, cache a table when hand cannot be drawn (plus, it can be modified in a same way as a table for invisible card areas)

## Additional Info:


SleepyG11:
Working with Lilo on additional optimizations, goes back to draft until all of them will be resolved

PR #1488, opened 2026-08-09
Arbitrary sets in `generate_card_ui`

wingedcatgirl:
For info queues and stuff. If a set name is unrecognized, the fallback behavior is now to treat it as a description, instead of doing nothing.

## Additional Info:


wingedcatgirl:
N.B. I'm like 95% sure there's a more elegant way to do this than the one I went with (this patch is happening in between two other patches that affect the same area and I don't even know how the order is determined) but I wasn't able to figure it what it was.

wingedcatgirl:
Wow! I don't know how git works. Let me try this again.

PR #1498, opened 2026-08-16
Allow Deck multiboxes & rework deck collection UI

lord-ruby:
Deck collection UI is reworked to better match the run select menu (partially because the old UI was being really annoying with resizing when multiboxes are present and because it has been on the to-do list for a while)
Multiboxes on decks should work the same as everything else

## Additional Info:

### DIFF
--- a/lovely/multi_box_descriptions.toml
+++ b/lovely/multi_box_descriptions.toml
@@ -209,7 +209,7 @@ position = 'before'
 match_indent = true
 payload = '''
 local config = {}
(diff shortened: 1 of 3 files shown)


lord-ruby:
crash involving clicking collection decks i forgot to account for, i am working on fixing it

MaxBoi342:
deck multiboxes dont seem to work with info queue, or at the very least using {T:[key]} 
image
image

MaxBoi342:
also the collections view doesnt work properly (deck descirptions dont show up on hover) on first load but works fine after switching a page

PR #1504, opened 2026-08-20
`SMODS.GUI.text_input`: multi-language text input

SleepyG11:
This PR adds functionality for creating text input via `SMODS.GUI.text_input(args)`, which have next features:
- Multi-language: support for any unicode symbols and special symbols from other languages
- Font: can specify which font to use for displaying text, especially important for unicode symbols
- Filters: can specify whitelist, blacklist and corpus
- Modify function: easy way to change inputted key conditionally, or for other side effects
- Key repeats: added functionality for inputting/removing multiple symbols while button is held (no more esc smashing)
- Based on vanilla text input, other arguments such as sizes, callbacks etc. stays the same

Balatro_YPKxr9VFMI

### Things to do:
- Decide which logic to use for non-multi-language special characters input
- LSP
- Mobile?
- Controller?

## Additional Info:

### DIFF
--- a/src/ui.lua
+++ b/src/ui.lua
@@ -3527,6 +3527,8 @@ G.FUNCS.smods_gui_text_input_key = function(args)
         hook_config.colour[3] = G.C.WHITE[3]
         ease_colour(hook_config.colour, temp_colour)
         G.CONTROLLER.text_input_hook = nil
+        love.keyboard.setKeyRepeat(hook_config.game_key_repeat)
+        love.keyboard.setTextInput(hook_config.game_text_input)
     elseif args.key == 'LEFT' then --Move cursor position to the left
         TRANSPOSE_TEXT_INPUT(-1)
     elseif args.key == 'RIGHT' then --Move cursor position to the right
@@ -3538,6 +3540,26 @@ G.FUNCS.smods_gui_text_input_key = function(args)
             pos = text.current_position+1
         }
         TRANSPOSE_TEXT_INPUT(1)
+    end
+end
+
+local old_select_text_input = G.FUNCS.select_text_input
+function G.FUNCS.select_text_input(...)
+    old_select_text_input(...)
+    local e = G.CONTROLLER.text_input_hook
+    if e and not e.REMOVED and e.config.ref_table and e.config.ref_table.smods_gui_input then
+        e.config.ref_table.game_key_repeat = love.keyboard.hasKeyRepeat()
+        e.config.ref_table.game_text_input = love.keyboard.hasTextInput()
+        love.keyboard.setKeyRepeat(true)
+        love.keyboard.setTextInput(true)
+        local old_remove = e.remove
+        function e:remove(...)
+            if G.CONTROLLER.text_input_hook == self then
+                love.keyboard.setKeyRepeat(e.config.ref_table.game_key_repeat)
+                love.keyboard.setTextInput(e.config.ref_table.game_text_input)
+            end
+            old_remove(self, ...)
+        end
     end
 end
 
@@ -3561,13 +3583,6 @@ function love.textinput(text, ...)
 		})
 	end
 	love_textinput(text, ...)
-end
-
-local old_load = love.load or function() end
-function love.load(...)
-    old_load(...)
-    love.keyboard.setKeyRepeat(true)
-    love.keyboard.setTextInput(true)
 end
 
 local old_key_button = G.FUNCS.key_button
--- a/src/ui.lua
+++ b/src/ui.lua
@@ -3535,3 +3535,9 @@ function love.textinput(text, ...)
 	end
 	love_textinput(text, ...)
 end
+
+local old_load = love.load or function() end
+function love.load(...)
+    old_load(...)
+    love.keyboard.setKeyRepeat(true)
+end
--- a/src/ui.lua
+++ b/src/ui.lua
@@ -3454,8 +3454,8 @@ G.FUNCS.smods_gui_text_input_key = function(args)
 
     -- Always take control inputs
     if not keymap[args.key] then
-        -- Ignore input from keypressed
-        if not args.textinput then return end
+        -- Reject input from keypressed unless it's from in-game screen keyboard
+        if args.keypressed and not SMODS.keypress_from_os_keyboard then return end
         -- Ignore input longer than 1 symbol
         if utf8Len(args.key) ~= 1 then return end
         -- Ignore non-printable keys if input is not multi-language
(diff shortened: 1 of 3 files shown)


src/ui.lua
```diff
@@ -3391,4 +3391,203 @@ G.FUNCS.HUD_blind_badge = function(e)
...
+    args.prompt_colour = args.prompt_colour or lighten(copy_table(args.colour), 0.4)
+    args.current_prompt_text = ''
+    args.id = args.id or "text_input"
+    args.multi_language = args.multi_language or false
+    args.smods_gui_input = true
+
+    local text = {ref_table = args.ref_table, ref_value = args.ref_value, letters = {}, current_position = utf8Len(args.ref_table[args.ref_value])}
+    local ui_letters = {}
+    for i = 1, args.max_length do
+        text.letters[i] = (args.ref_table[args.ref_value] and (utf8CharAt(args.ref_table[args.ref_value], i) or '')) or ''
```
WilsontheWolf, line 3430: This crashes if the ref_value is not a string (such as a number), when the old one doesn't

src/ui.lua
```diff
@@ -3391,4 +3391,203 @@ G.FUNCS.HUD_blind_badge = function(e)
...
+    args.h = args.h or 0.7
+    args.text_scale = args.text_scale or 0.4
+    args.max_length = args.max_length or 16
+    args.all_caps = args.all_caps or false
+    args.prompt_text = args.prompt_text or localize('k_enter_text')
+    args.prompt_colour = args.prompt_colour or lighten(copy_table(args.colour), 0.4)
+    args.current_prompt_text = ''
+    args.id = args.id or "text_input"
+    args.multi_language = args.multi_language or false
+    args.smods_gui_input = true
```
WilsontheWolf, line 3421: I'd have to do more through testing, but I believe that it makes sense to make the defaults for this more flexible (My inital thoughts are multi_language = true, max_lenght = inf or maybe 256) and then have the old text input just call this one with it's defaults instead of the new defaults. I don't really see a need to keep both inputs around

src/ui.lua
```diff
@@ -3391,4 +3391,203 @@ G.FUNCS.HUD_blind_badge = function(e)
...
+    local hook = G.CONTROLLER.text_input_hook
+    local hook_config = hook.config.ref_table
+    local text = hook_config.text
+
+    -- Always take control inputs
+    if not keymap[args.key] then
+        -- Reject input from keypressed unless it's from in-game screen keyboard
+        if args.keypressed and not SMODS.keypress_from_os_keyboard then return end
+        -- Ignore input longer than 1 symbol
+        if utf8Len(args.key) ~= 1 then return end
```
WilsontheWolf, line 3474: Why?

src/ui.lua
```diff
@@ -3391,4 +3391,203 @@ G.FUNCS.HUD_blind_badge = function(e)
...
+    return old_text_input_key(args, ...)
+end
+
+local love_textinput = love.textinput or function() end
+function love.textinput(text, ...)
+	local e = G.CONTROLLER.text_input_hook
+	if e and not e.REMOVED and e.config.ref_table and e.config.ref_table.smods_gui_input then
+		G.FUNCS.smods_gui_text_input_key({
+			key = text,
+			caps = G.CONTROLLER.held_keys["lshift"] or G.CONTROLLER.held_keys["rshift"],
```
WilsontheWolf, line 3595: I thought love.textinput handled caps for us

src/ui.lua
```diff
@@ -3391,4 +3391,203 @@ G.FUNCS.HUD_blind_badge = function(e)
...
+    }
+
+    local hook = G.CONTROLLER.text_input_hook
+    local hook_config = hook.config.ref_table
+    local text = hook_config.text
+
+    -- Always take control inputs
+    if not keymap[args.key] then
+        -- Reject input from keypressed unless it's from in-game screen keyboard
+        if args.keypressed and not SMODS.keypress_from_os_keyboard then return end
```
WilsontheWolf, line 3472: When would this get a keypress that is not from the keyboard?

src/utils.lua
```diff
@@ -2,6 +2,29 @@
...
+
+function utf8Len(...)
+    return utf8.len(...)
+end
+function utf8Offset(...)
+    return utf8.offset(...)
+end
+function utf8Codepoint(...)
+    return utf8.codepoint(...)
+end
```
WilsontheWolf: Why make these globals instead of just requiring the module where needed?

WilsontheWolf: If you do have a really good reason to make these globals, why not like 

```suggestion
local utf8 = require("utf8")

utf8Len = utf8.len
utf8Offset = utf8.offset
utf8Codepoint = utf8.codepoint
```

WilsontheWolf (changes requested):

src/ui.lua
SleepyG11: `G.FUNCS.smods_gui_text_input_key` processes `love.textinput` for inputting actual letters, and `love.keypressed` for control inputs, such as left and right arrows, backspace and enter events. If we remove this length check, we will start inputting symbols such as Ctrl, Alt, Caps Lock and other.
Tho moving this check after calling `hook_config.func()` makes more sense.

src/ui.lua
SleepyG11: Maybe it is, I was just following vanilla caps handling approach.
Maybe makes sense call string.lower when it's not caps.

src/ui.lua
SleepyG11: `max_length = inf` will explode your pc because it creates new text element for each letter which can be fit in this limit. Thunk moment. Changing that will require complete rebuilding of how text input as ui element operates. Not easy task to do.
`multi_language = false` by default because this element intented to be "better text input with multi-language feature", not "multi-language text input", despite PR name saying otherwise :droll:

src/ui.lua
SleepyG11: Idea is: if function called from `G.CONTROLLER:key_press_update`, process it only if it was either special key (left/right arrow, back, enter), or from screen keyboard to allow input from it. Otherwise we would need manually call `love.textinput(key)` to make screen keyboard work.

src/utils.lua
SleepyG11: utf8 functions needed in multiple files:
- [SMODS _ "src/ui.lua"]
- functions/button_callbacks.lua
- functions/UI_definitions.lua

Doing `utf8 = require('utf8')` once during game loading leads to crashes, requiring module on every function call will require more patches + doesn't feel like good tactic. So I decided to define functions I need, plus can make my own such as `utf8CharAt` in same style.

1-line defining makes sense tho.

src/ui.lua
SleepyG11: I guess I'll stringify it then 💀

src/ui.lua
SleepyG11: Vanilla input should be kept because of backwards compatibility. There's a lot of patches from various mods that applied to `G.FUNCS.text_input` specifically to make it behave as they need; breaking it is not what I like to do.
Plus, doing that will change behaviour of already existing vanilla inputs, such as profile name and seed. Dont want mess around with them too.

PR #1505, opened 2026-08-20
Auto-add Stickers to loc_colours

VMan-2002:
Automatically add loc_colours entries corresponding to stickers

## Additional Info:


nh6574:
I feel like this is somewhat arbitrary since other objects can also have custom colours for badges and other custom colours in general. Vanilla also uses `{C:attention}` for stickers so I wouldn't think a dev would necessarily expect it to be added to the formatting colours like the ones SMODS currently adds automatically. 

What worries me is that adding more colours could lead to name conflicts, especially since stickers don't have a class prefix.

PR #1512, opened 2026-08-22
Card:start_materialize Custom Set Color Fix

DJKart22:
Allows custom sets to properly materialize as the designated secondary color opposed from the default green

## Additional Info:


nh6574:
Did this happen on latest dev? Because that patch was already meant to solve that and I don't see how this modification solves the issue. In fact, this seems like it might crash if there's no color defined.

DJKart22:
It was not on latest dev, and it does seem to be working properly in that version. No change is needed.

PR #1531, opened 2026-09-03
Fix `SMODS.card_to_image`

wingedcatgirl:
Started working on making it not choke on cards of sizes other than 71x95 exactly
Still some hiccups, it rendered FaC's green pirate hat (710x950) as a 71x95 image stuck in the corner of a 710x950 canvas and idk why. It does mostly work but these hiccups confuse me. Thus the draft status to see if anyone else can figure it out.

## Additional Info:


AllUniversal:
My guess for the green hat issue is that you need to multiply the scale by `px/71`. Maybe try it with a 35x47 image as well!

wingedcatgirl:
I still don't know how to handle the case of display size ratios not being square but y'know what. I'm not sure that's even real

It works with the pirate hat now so I'm taking this as correct enough

PR #1537, opened 2026-09-07
Fix crash from negative hand levels

The-Arx:
The cause of the bug is this snippet from `update_hand_text` that doesn't account for negative levels, which makes the colour nil causing a crash later on in `UIElement:draw_self`
```lua
if type(vals.level) == 'number' then 
    G.hand_text_area.hand_level.config.colour = G.C.HAND_LEVELS[math.min(vals.level, 7)]
else
    G.hand_text_area.hand_level.config.colour = G.C.HAND_LEVELS[1]
end
```
The fix I implemented is to add a nil check after this if statement and fallback to G.C.RED (the color for level 0)
```lua
if not G.hand_text_area.hand_level.config.colour then
    G.hand_text_area.hand_level.config.colour = G.C.HAND_LEVELS[0]
end
```
A lot of mods use the pattern `G.hand_text_area.hand_level.config.colour = G.C.HAND_LEVELS[math.min(vals.level, 7)]` so I avoided changing that line to not break compatibility.

I tested that it works but not the compatibility with other mods, although I have looked through most of the references to this code on github search and haven't seen any mods or patches that it breaks.

~~The only mod it might break is suikalatro because they [match against the entire function](https://github.com/pi-cubed-cat/suikalatro/blob/main/lovely/play_hand_fixes.toml#L101) to remove a few lines.~~ (I just tested and the patch I linked already fails to apply on the latest steamodded version)

## Additional Info:

### DIFF
--- a/lovely/fixes.toml
+++ b/lovely/fixes.toml
@@ -973,3 +973,22 @@ G.PROFILES[G.SETTINGS.profile].all_unlocked = true
 payload = '''
 EMPTY(G.P_LOCKED)
 '''
+
+# Fix negative hand levels causing a crash
+[[patches]]
+[patches.pattern]
+target = 'functions/common_events.lua'
+match_indent = true
+position = 'after'
+pattern = '''
+if type(vals.level) == 'number' then 
+    G.hand_text_area.hand_level.config.colour = G.C.HAND_LEVELS[math.min(vals.level, 7)]
+else
+    G.hand_text_area.hand_level.config.colour = G.C.HAND_LEVELS[1]
+end
+'''
+payload = '''
+if not G.hand_text_area.hand_level.config.colour then
+    G.hand_text_area.hand_level.config.colour = G.C.HAND_LEVELS[0]
+end
+'''


MetaNite64:
I don't think negative hand levels are ever intended to be accessible in the first place? What mod are you playing/making that lets this happen?

The-Arx:
For example in multiplayer mod the game will crash after too many asteroids

The-Arx:
Many mods have card level down effects (many are limited to level 0 or 1 but that might be to prevent this crash in the first place): https://github.com/search?q=%2Flevel_up_hand%28%5B%5E%29%5D%2B%2C+-1%29%2F+language%3Alua&type=code

Some mods even added their own fixes (that this fix wouldn't interfere with since it is adds a null check after):
https://github.com/connorwillchris/Degeneracy/blob/main/lovely.toml#L67

nh6574:
Do you have steps to recreate the crash? Because from my testing the colors do break but it doesn't crash for me.

Edit: Ok, it seems to happen for only specific value changes. I'll like to see the exact cause first before merging

PR #1544, opened 2026-09-09
Added `SMODS.CalculationEffect`

AllUniversal:
+/*Title, refactored `SMODS.calculate_individual_effect()` accordingly.
-> This allows more easily adding new calculate effects, e.g. `x_dollars`.

! Edit: This now includes breaking changes for `SMODS.Scoring_Parameter`; It no longer uses a `calc_effect` field, instead it creates any number of `SMODS.CalculationEffect`s as defined in its new `calculation_effects` field when injected.

## Additional Info:


HuyTheKiller:
This is actually so neat. Having to do so via many hooks isn't really dev-friendly.

src/game_objects/calculation_effect.lua
```diff
@@ -0,0 +1,541 @@
...
+            return true
+        end
+    end
+}
+
+SMODS.CalculationEffect {
+    key = "message",
+    order = 60,
+    func = function (self, effect, scored_card, key, amount, from_edition)
+        if not SMODS.no_ressolve then
```
Aurelius7309: `no_resolve`

Issue #760, opened 2025-06-12
quantum wildcards causing massive lag


SadCubeArt:
Heya, i tried to make a joker that makes clubs act like wild cards sing quantum enhancement feature, and while it seems to work the effect causes massive lag! i attatched the code of the joker, maybe you can figure out whats the issue/ solution 
[rorschach.txt](https://github.com/user-attachments/files/20712668/rorschach.txt)

Issue #601, opened 2025-04-01
Quantum Enhancements' perfomance


VictinZero:
Based on reports from various modders, the current state of quantum enhancements introduces lag up to inducing crashes as the game runs out of memory. This is in contrast with the feature seemingly performing well previously.

This has been discussed on Discord and this GitHub issue is to register the problem. It is unclear when the current issues were introduced or more specific details, although they do apply over multiple mods.

SidOfc:
I'm not sure why this was closed however I seem to be running into some issues with this as well, I initially posted this at Pokermon since I figured it may have to do with how the mechanics I was using were implemented there but then sonfive responded regarding this relating to Quantum Enhancements and how smods handles them.

For reference, [this is my initial post](https://github.com/InertSteak/Pokermon/issues/716) (includes more details and full list of mods).

I don't want to create any duplication for maintainers so I'll leave this here as a comment on an existing issue for a bit before I open a new issue about it.

Aurelius7309:
This was meant to be closed in reference to https://github.com/Steamodded/smods/commit/cd295e9ea7087a688a3eb8bd9e5e050e3335c6c8, which did fix some issues that were present at the time but not the general performance of the feature as a whole. Reopening this as it's not tracked elsewhere

SidOfc:
Cheers for reopening, I must say I'm amazed at the response times both from Pokermon and smods, if there's anything I can do let me know! I can provide my mods folder or save file if instructed how to do so or tinker around with some debug mods.

Thanks again for reopening! 👍

PR #1552, opened 2026-09-14
[Discussion] Avoid clearing enhancement cache every update

caner-cetin:
## Discussion status

This is intentionally a draft. The performance improvement is dramatic in a heavily modded run, but the correctness implications for dynamic quantum enhancements need maintainer input.

## Problem

`lovely/enhancement.toml` injects `SMODS.enh_cache:clear()` after `modulate_sound(dt)`, clearing the complete enhancement cache on every game update.

A 37-second capture on an M4 Mac recorded:

- 596 global enhancement-cache clears
- 12,630 empty base-card cache repopulations
- 13,625 full `check_enhancement` calculations
- 53,210 calls to `SMODS.get_enhancements`
- about 30.9 seconds of inclusive `get_enhancements` time

Every cache miss calls `SMODS.calculate_context({ check_enhancement = true, ... })`, which evaluates the loaded card areas. Large modded card areas therefore magnify the cost substantially.

## Experimental change

Remove only the unconditional per-update cache clear. Existing targeted invalidations remain in place for card ability, base, seal, edition, sticker, card-add, and voucher changes.

In the same affected save, the previously delayed calculation became effectively instant after installing this change.

## Prior-art investigation

This per-update invalidation does **not** appear to be an accidental leftover:

- [#408](https://github.com/Steamodded/smods/pull/408) introduced the initial enhancement-cache prototype and explicitly described it as an unclean implementation intended to start discussion.
- [#412](https://github.com/Steamodded/smods/pull/412) replaced that prototype with the implementation currently in SMODS. Its discussion considered a roughly one-second cache lifetime, recalculating only during scoring/hover, and invalidating cards at specific mutation events. It also identified stale results for consumers such as score-preview/Joker-display mods as a risk. The merged implementation retained the per-update reset, including after its final `Safer clear conditions` commit.
- [#601](https://github.com/Steamodded/smods/issues/601) remains the upstream tracking issue for general quantum-enhancement performance, including reports of lag and out-of-memory crashes across multiple mods. A maintainer clarified that `cd295e9` fixed recursion-related issues but not the feature's general performance.
- [#760](https://github.com/Steamodded/smods/issues/760) independently reports severe lag from a quantum wildcard effect.
- [#945](https://github.com/Steamodded/smods/pull/945) fixed infinite getter-context recursion involving `check_enhancement`, illustrating why cache/context behavior has non-obvious cross-mod correctness cases.
- [#1322](https://github.com/Steamodded/smods/pull/1322) optimized quantum `set_ability` work (reported 8 seconds down to 3 seconds for its stress case), while documenting compatibility tradeoffs.
- [#1351](https://github.com/Steamodded/smods/pull/1351), the open Generalized Quantum API PR, proposes a much broader quantum-field cache and explicit/automatic cache updates. Its discussion is still working through performance, per-frame cache lifetime, and stateful quantum-object semantics.

Therefore this PR is not claiming that deleting the reset is a complete cache-coherency design. It provides a small, measured reproduction showing that the conservative reset can dominate runtime, and asks whether the safe fix should be a dirty-generation/event invalidation scheme or should instead be handled by the broader work in #1351.


Aurelius7309:
The existing quantum enhancements implementation is flawed in a lot of ways, and I wouldn't want to concern myself with trying to fix it in any meaningful way. The current behavior of clearing cache on update was the result of some inconsistencies, and it's really a band-aid fix to a broader problem with the implementation itself. Thus, this issue is only to be considered within the context of #1351, and changing the reset behavior of what's there will not be considered.

SleepyG11:
As quick optimization which can be done, is removing double cache read inside `SMODS.get_enhancements()` on cache hit, lines 1033 and 1054.

https://github.com/Steamodded/smods/blob/c0acce9f2c2af9c5876450407ecff18d45a881f8/src/utils.lua#L1029-L1055

```lua
local ret = SMODS.enh_cache:read(card, extra_only)
if not ret then
  -- quantum stuff calculation
  ret = SMODS.enh_cache:read(card, extra_only)
end
return ret
```

Additionally, we can optimize cache read when `extra_only` argument is not present (99% cases).

https://github.com/Steamodded/smods/blob/c0acce9f2c2af9c5876450407ecff18d45a881f8/src/utils.lua#L1061-L1066

```lua
read = function(self, key, extra_only)
    local ret = self.data[key]
    if extra_only and ret then
        ret = copy_table(ret)
        ret[key.config.center.key] = nil
    end
    return ret
end
```

Also, isn't this part should call quantum set ability and not regular one?
https://github.com/Steamodded/smods/blob/c0acce9f2c2af9c5876450407ecff18d45a881f8/lovely/enhancement.toml#L162-L187

Issue #1545, opened 2026-09-10
Crash in has_playing_card_property when unlock popup renders with quantum_enhancements enabled


tibernero:
With quantum_enhancements on, during a run, an unlock notification crashes the game:

```
[SMODS _ "src/utils.lua"]:1131: attempt to index a nil value
```

`create_UIBox_card_unlock` builds its locked placeholder card (center key `j_locked`), `get_enhancements` seeds that key into the enhancement list, and the loop in `has_playing_card_property` indexes `G.P_CENTERS["j_locked"]`, which is nil. The `G.OVERLAY_MENU` guard added in 4bcdd83 doesn't cover this: the card is created before the overlay is set, so `get_enhancements` still takes the quantum path.

Changing the loop to `(G.P_CENTERS[k] or {})[key]` (the idiom the edition/seal checks below it already use) fixes it — confirmed in-game. Happy to PR the one-liner if you want.

nh6574:
I can't manage to replicate this at all. Do you have any other information, like the full crash log?

Edit: Managed to. It seems like it doesn't happen for every unlock.

PR #1555, opened 2026-09-14
Fix invalid quantum enhancement center lookup

caner-cetin:
## Summary

Prevent unregistered native center keys from being added to the quantum enhancement table.

## Cause

With `quantum_enhancements` enabled, `SMODS.get_enhancements(card)` adds every native center key other than `c_base` before processing context-provided enhancements. The context-provided keys are validated against `G.P_CENTERS`, but the native key was not.

Vanilla Balatro's `create_UIBox_card_unlock` creates a valid display card using the placeholder center object `G.j_locked`. That center exists, but `G.P_CENTERS.j_locked` is nil. The unvalidated insertion therefore produced `{ j_locked = true }`, which later caused `SMODS.has_playing_card_property` to index `G.P_CENTERS.j_locked.replace_base_card`.

## Fix

Apply the same `G.P_CENTERS` validation to the native center key that is already applied to context-provided enhancement keys. This prevents non-enhancement placeholder centers from entering the enhancement cache.

## Behavior

Registered centers are handled identically. Placeholder or otherwise unregistered centers cannot represent registered enhancements and are omitted from the enhancement table.

## Observed stack

`create_UIBox_card_unlock` -> `Card(..., G.j_locked)` -> `Card:set_ability` -> `Card:should_hide_front` -> `SMODS.has_playing_card_property`

The captured locals showed `card.config.center.key == "j_locked"`, `SMODS.get_enhancements(card) == { j_locked = true }`, and `G.P_CENTERS.j_locked == nil`.


SleepyG11:
This check is redundant because they're already included inside `SMODS.get_enhancements()`; you cannot create a card without center, otherwise game will crash.
https://github.com/Steamodded/smods/blob/c0acce9f2c2af9c5876450407ecff18d45a881f8/src/utils.lua#L1041-L1051

caner-cetin:
The card does have a center; the issue is that the center object is not necessarily registered in `G.P_CENTERS`.

The crash path was `create_UIBox_card_unlock` -> `Card(..., G.j_locked)` -> `Card:set_ability` -> `Card:should_hide_front` -> `SMODS.has_playing_card_property`. The captured locals showed the center as `{ key = j_locked, name = Locked, set = Joker, unlocked = false, ... }`, while `G.P_CENTERS.j_locked` was nil. `create_UIBox_card_unlock` is vanilla Balatro code, so there is no corresponding file in this repository to link.

The linked filtering only applies to keys returned by `calculate_context()`:
https://github.com/Steamodded/smods/blob/c0acce9f2c2af9c5876450407ecff18d45a881f8/src/utils.lua#L1044-L1051

The card's native center key is inserted before that, without checking `G.P_CENTERS`:
https://github.com/Steamodded/smods/blob/c0acce9f2c2af9c5876450407ecff18d45a881f8/src/utils.lua#L1039-L1043

That leaves `{ j_locked = true }` for this valid placeholder center. The unchecked property lookup is here:
https://github.com/Steamodded/smods/blob/c0acce9f2c2af9c5876450407ecff18d45a881f8/src/utils.lua#L1133-L1142

So this is not a card without a center; it is a card whose valid vanilla placeholder center is outside `G.P_CENTERS`.

caner-cetin:
I've moved the validation to the native-center insertion in `SMODS.get_enhancements()`, alongside the existing validation for context-provided keys. The final diff now prevents `j_locked` from entering the enhancement table rather than guarding the later consumer. I also corrected the PR description to use the captured `create_UIBox_card_unlock` path.

SleepyG11:
I see, makes sense. Lgfm.

SleepyG11:
Fixes #1545

PR #1558, opened 2026-09-15
`SMODS.add_round_eval_row`

nh6574:
Replaces `add_round_eval_row` with a nicer more flexible function.

Currently almost complete and I would appreciate testing and review to see that I didn't miss anything.

Still to do:
  - Replace SMODS uses of `add_round_eval_row` for `SMODS.add_round_eval_row` and interface calc_dollar_bonus returns with it. Vanilla uses would remain since this function removes some hardcoded stuff from it.
  - `args.card`/`args.object` which allow the function to automatically get the name from the object currently don't work since I plan to make it easier to get prototype objects and their localization placement in another PR soon-ish. If we want to merge this before that we can just copy the code that handles calc_dollar_bonus.

## Additional Info:


nh6574:
(waiting on https://github.com/Steamodded/smods/pull/1559 and probably a later PR to get the localized names of objects easier after that one)

PR #1559, opened 2026-09-15
`obj.prototype`

nh6574:
Adds an easy want to find prototype object from game objects (i.e. the SMODS.Center of a card, the SMODS.Rank of a card, SMODS.Tag of a tag, or the vanilla equivalents)

Added:
  - `card.prototype` == `card.config.center`
  - `blind.prototype` == `blind.config.blind`
  - `deck.prototype` == `deck.effect.center`
  - `tag.prototype` == `G.P_TAGS[tag.key]`
  - `tag_sprite.prototype` == `G.P_TAGS[tag_sprite.config.tag.key]`
  - `G.stake_prototype` == `G.P_CENTER_POOLS.Stake[G.GAME.stake]`
  - `card.rank_prototype` == `SMODS.Ranks[card.base.value]` (`nil` if none)
  - `card.suit_prototype` == `SMODS.Suits[card.base.suit]` (`nil` if none)
  - `card.edition_prototype` == `G.P_CENTERS[card.edition.key]` (`nil` if none)
  - `card.seal_prototype` == `G.P_SEALS[card.seal]` (`nil` if none)

Additionally I added `card.stickers` as a key-indexed table with all of the cards stickers that gets saved to the card (except when it gets applied directly and not using one of the functions, which means `card.pinned` will practically never be here... I can solve that with a bunch of patches if it's wanted. Also DebugPlus doesn't use the functions either.). 

`card.sticker_prototypes` is similar but it contains each prototype object and doesn't get saved. Maybe it can be just one table that just gets the prototypes removed on save.

Not changed:
  - `G.GAME.challenge_tab`: The challenge's prototype.
  - `card.config.card`: The card object prototype. This one I don't know where it would be used and it's not that hard to access.

Also while the term prototype has been used before (I think by some of my other PRs) this is a good time to standardize it so if we don't like this one we can change it to something else.

## Additional Info:


SleepyG11:
What if `card.config.center == G.j_locked`? It's not inside `G.P_CENTERS`

nh6574:
`card.prototype` is also `G.j_locked` in that case.

nh6574:
Ok, scrap that (after some careful consideration in the shadows of the SMODS discord)

This is how it works now: Each SMODS class with a corresponding game object now has a `get_prototype_object` static method.

For example: 
  - `SMODS.Center.get_prototype_object(card)` will get `card.config.center`
  - `SMODS.Edition.get_prototype_object(card)` will get `G.P_CENTERS[card.edition.key]` (`nil` if none)
  - `SMODS.Blind.get_prototype_object(blind)` will get `blind.config.blind`

To keep the practicality of the earlier implementation each game object gets a `prototype` function (`Card:prototype()`, `Blind:prototype()`, `Tag:prototype()` and `Back:prototype()`).

Some notes on this:
  - `Card:prototype(obj_type)` gets the center if `obj_type` is `nil`, otherwise it gets the corresponding object mapped at `SMODS.card_prototype_map`. E.g. `card:prototype("rank") == SMODS.Rank.get_prototype_object(card)` . 
     - Valid arguments: `"center"`, `"enhancement"` (same as `center` internally), `"edition"`, `seal`, `sticker`, `stickers` (plural alias for the reasons below, but it acts the same), `rank`, `suit`, `rarity` and `card` (`card.config.card`)
  - Stickers are of course special since there can be multiple, so `card:prototype("sticker")` returns a similar key-indexed table to what I had before. Also `SMODS.Sticker.get_prototype_object` is now an alias of `SMODS.Sticker.get_stickers`, basically. The reason I didn't add `Card:get_stickers()` is because that would probably fall into the quantum jurisdiction while this is explicitly to get the base sticker prototypes, so it belongs in `SMODS.Sticker` (imo). Similar to how `card:prototype()` doesn't return quantum enhancements.
  - While it might now fall out of purview of this PR, I also added `SMODS.Stake.get_applied()` and `SMODS.Challenge.get_current()` to get the prototypes of the stakes and current challenge respectively.

nh6574:
Would it be useful to add an `obj:get_key()` to this?

PR #1560, opened 2026-09-16
Dynamic RunSelect Selection Limit

the-Astra:
Allows the selection_limit property of a RunSelect page to be a function, allowing dynamic selection limits based on certain conditions

## Additional Info:


WilsontheWolf (changes requested):
I'm annoying

Eremel (changes requested):
I think a function `SMODS.RunSelect.Functions.get_selection_limit(page_def)` would be preferable to repeating the same snippet of code multiple times throughout the code base

